PR #26, opened 2020-11-30
Switch to artifact URLs for artifacts

micahhausler:
Update to use URI instead of Path, this requires less context for the user to add the domain


brandond:
Thanks for fixing this!

PR #47, opened 2020-12-08
Set session name on prow cp pods

micahhausler:
*Issue #, if available:*

*Description of changes:*

* Adds a session name to help debug which IAM role sessions are associated with which pods 
* Ensure that aws-iam-authenticator is using the regional endpiont


vivek-koppuru (approved):
/lgtm
/approve

PR #51, opened 2020-12-09
Set the path variable to use helm

bnrjee:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/projects/gomods/athens/upload-athens-artifacts.sh
+++ b/projects/gomods/athens/upload-athens-artifacts.sh
@@ -21,15 +21,17 @@ err_report() {
 }
 trap 'err_report $LINENO' ERR
 
+MAKE_ROOT="$(pwd -P)"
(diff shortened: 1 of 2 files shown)


projects/gomods/athens/upload-athens-artifacts.sh
```diff
@@ -21,15 +21,17 @@ err_report() {
 }
 trap 'err_report $LINENO' ERR
 
+MAKE_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)"
```
vivek-koppuru, line 24: This isn't the root of the Makefile right? I think doing a `pwd -P` might be enough here, and then changing the below lines to add an addition `../`

vivek-koppuru (changes requested):

vivek-koppuru (approved):
/lgtm
/approve

PR #58, opened 2020-12-16
Add token-expiration and sts-regional-endpoint flags

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*

Adding parameters to webhook so that default token expiration is 3600 and sts regional endpoint is set to true.

Not modifying version as we haven't deployed this yet.


micahhausler (approved):
/lgtm
/approve

PR #60, opened 2020-12-17
Rename webhook so it doesn't override eks webhook

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*

Rename pod identity webhook to avoid replacing configuration by eks. Also fixed quotes in job script to be able to use the env var in the string, and increased sleep and loop count.


micahhausler (approved):
/lgtm
/approve

PR #62, opened 2020-12-18
Separate webhook configuration naming and add uninstall script to cleanly remove resources

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*

Added a new value specifically for the mutatingwebhookconfiguration resource. Also added a pre-delete hook that would remove the webhook and secrets before going forward with the rest of the helm deletion. 

Also modified the loop in install.sh to sleep of 5 seconds in between each check for pending csr's for a total of 20 seconds. I observed that for all 3 clusters, it is able to detect the new csr in the first 15 seconds anyways, and since we want to be mindful of the helm execution timeout, 100 seconds is more than enough.


micahhausler (approved):
/approve
/lgtm

PR #64, opened 2020-12-18
Decrease loop for install.sh in pod-identity-webhook helm chart

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*

The loop for the job that deploys in our fargate cluster is too long, causing a timeout when helm is waiting for the hook to finish. I didn't encounter this issue on my devstack, but the csr's got approved within the first 2 loops, so 10 should definitely be enough for us.


vignesh-goutham (commented):
/approve
/lgtm

vignesh-goutham (approved):
/lgtm
/approve

PR #71, opened 2020-12-23
Add bash 4.3 to builder base to support kubernetes make test

jaxesn:
Running make test from the upstream kubernetes source requires bash version of 4.3 or higher.  Amazon linux ships with 4.2 and there doesn't seem to be a way to get it without building it from source.

I have built this docker image locally and built kubernetes and run the unit tests with it.

light reading:
https://github.com/kubernetes/kubernetes/issues/82605
https://github.com/kubernetes/kubernetes/issues/78702


micahhausler (approved):
/lgtm
/approve

PR #88, opened 2021-01-12
Remove minor version from our paths for golang

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*

This is a continuation of https://github.com/aws/eks-distro-build-tooling/pull/86 and related to https://github.com/aws/eks-distro/pull/114

Since we don't need to know the minor version and only will use golang binaries based on the major version, such as "1.15", I am modifying the path that we are storing them in so for the related PR, we won't have to hardcode the versions in our script.

Also modified to just using a function instead so I can make use of local variables, and after some discussions on functions vs loops earlier.

### DIFF
--- a/builder-base/install.sh
+++ b/builder-base/install.sh
@@ -87,9 +87,16 @@ mkdir -p /go/src /go/bin /go/pkg /go/src/github.com/aws/eks-distro
 export GOPATH=/go
 export PATH=${GOPATH}/bin/:$PATH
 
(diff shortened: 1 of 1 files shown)


TerryHowe:
I'm not following why that test is failing

vivek-koppuru:
> I'm not following why that test is failing

This PR should fix it: https://github.com/aws/eks-distro-build-tooling/pull/90

abhay-krishna (approved):

PR #90, opened 2021-01-13
Use PR branch ref SHA for presubmits instead of base ref

abhay-krishna:
Using the PULL_BASE_SHA environment variable for simulating the changes in the presubmit jobs could cause issued since it would essentially be trying to replace the existing tag with the same new tag value. This causes `git add` to not add any files, hence `git commit` fails with exit code 1 since there are no files to commit.

Hence for presubmits, we could use `PULL_PULL_SHA`, a job environment variable from the presubmit CI environment which refers to the SHA of the head commit of the PR against which the job runs.


vivek-koppuru (approved):
/lgtm
/approve

PR #96, opened 2021-01-23
Install AWSCLIv2 on builder-base image

abhay-krishna:
Though AWS CLI version 2 is generally available, the builder-base image still uses version 1 of the CLI since it is downloaded via `yum` and the latest available version in the `yum` repositories is `1.18.147`. Hence we should install version 2 using [these instructions](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2-linux.html#cliv2-linux-install).

### DIFF
--- a/builder-base/install.sh
+++ b/builder-base/install.sh
@@ -30,7 +30,6 @@ yum update -y
 
 amazon-linux-extras enable docker
 yum install -y \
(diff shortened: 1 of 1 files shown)


vivek-koppuru (approved):
/lgtm

PR #98, opened 2021-01-27
Add support for empty repositories list in CP values

abhinavmpandey08:
*Issue #, if available:*

*Description of changes:*
Add support for empty repositories list in CP values. 
Currently, if we pass in empty repositories to prow-control-plane helm chart, it will create invalid ConfigMaps. This PR solves that issue.

### DIFF
--- a/helm-charts/stable/prow-control-plane/templates/config-ConfigMap.yaml
+++ b/helm-charts/stable/prow-control-plane/templates/config-ConfigMap.yaml
@@ -83,7 +83,6 @@ data:
             entrypoint: {{ .Values.utility_images.entrypoint }}
             initupload: {{ .Values.utility_images.initupload }}
             sidecar: {{ .Values.utility_images.sidecar }}
-
     tide:
       queries:
       - labels:
@@ -94,12 +93,15 @@ data:
         - do-not-merge/hold
         - do-not-merge/work-in-progress
         - do-not-merge/invalid-owners-file
-        repos:
-        {{-  range $repo := .Values.repositories }}
-          - {{ $repo.org }}/{{ $repo.name }}
-        {{- end }}
+        repos: 
+        {{- if eq (len .Values.repositories) 0 }} []
+        {{- else }}
+          {{-  range $repo := .Values.repositories }}
+        - {{ $repo.org }}/{{ $repo.name }}
+          {{- end }}
       merge_method:
-      {{-  range $repo := .Values.repositories }}
+        {{-  range $repo := .Values.repositories }}
         {{ $repo.org }}/{{ $repo.name }}: squash
-      {{- end }}
+        {{- end }}
+        {{- end }}
     decorate_all_jobs: true
--- a/helm-charts/stable/prow-control-plane/templates/plugins-ConfigMap.yaml
+++ b/helm-charts/stable/prow-control-plane/templates/plugins-ConfigMap.yaml
@@ -18,7 +18,9 @@ metadata:
   name: plugins
 data:
   plugins.yaml: |
-    plugins:
+    plugins: 
+    {{- if eq (len .Values.repositories) 0 }} []
+    {{- else }}
       {{- range $repo := .Values.repositories }}
       {{ $repo.org }}/{{ $repo.name }}:
       - approve
@@ -39,6 +41,7 @@ data:
       - {{ $extra }}
       {{- end }}
       {{- end }}
+    {{- end }}
     config_updater:
       maps:
 {{ toYaml .Values.plugins.configUpdaterMaps | indent 8 }}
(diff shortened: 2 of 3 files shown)


helm-charts/stable/prow-control-plane/templates/config-ConfigMap.yaml
```diff
@@ -94,6 +94,10 @@ data:
         - do-not-merge/hold
         - do-not-merge/work-in-progress
         - do-not-merge/invalid-owners-file
+        {{- if eq (len .Values.repositories) 0 }}
+        repos: []
+      merge_method: {}
+        {{- else }}
```
vivek-koppuru, line 100: The indentation looks off here right?

helm-charts/stable/prow-control-plane/templates/config-ConfigMap.yaml
```diff
@@ -102,4 +106,5 @@ data:
       {{-  range $repo := .Values.repositories }}
         {{ $repo.org }}/{{ $repo.name }}: squash
       {{- end }}
+        {{- end }}
```
vivek-koppuru, line 109: Same here

helm-charts/stable/prow-control-plane/templates/config-ConfigMap.yaml
vivek-koppuru: Also instead of the conditionals on the key, can we do it for the values themselves. That way, it can input an empty list to `repos` instead of keeping track of the `repos` key in both the if and else condition.

vivek-koppuru (changes requested):

helm-charts/stable/prow-control-plane/templates/config-ConfigMap.yaml
abhinavmpandey08: The indentation looks weird because line 98 and 99 are on different indentation levels. Lines 97, 100 and 109 are all part of the one if-else-end block which is why I indented them the same amount. I wasn't really sure how else to indent them. Maybe breaking down the if-else into two will make it more readable. Something like this

```
        {{- if eq (len .Values.repositories) 0 }}
        repos: []
        {{- else }}
        repos:
        {{-  range $repo := .Values.repositories }}
          - {{ $repo.org }}/{{ $repo.name }}
        {{- end }}
        {{- end }}
      {{- if eq (len .Values.repositories) 0 }}  
      merge_method: {}
      {{- else }}
      merge_method:
      {{-  range $repo := .Values.repositories }}
        {{ $repo.org }}/{{ $repo.name }}: squash
      {{- end }}
      {{- end }}
    decorate_all_jobs: true
```

micahhausler: +1 to breaking it into 2 if/els blocks for readability

helm-charts/stable/prow-control-plane/templates/config-ConfigMap.yaml
```diff
@@ -83,7 +83,7 @@ data:
             entrypoint: {{ .Values.utility_images.entrypoint }}
             initupload: {{ .Values.utility_images.initupload }}
             sidecar: {{ .Values.utility_images.sidecar }}
-
+
```
vivek-koppuru, line 86: NIT: remove this extra space

helm-charts/stable/prow-control-plane/templates/plugins-ConfigMap.yaml
```diff
@@ -18,6 +18,9 @@ metadata:
   name: plugins
 data:
   plugins.yaml: |
+    {{- if eq (len .Values.repositories) 0 }}
+    plugins: []
```
vivek-koppuru, line 22: Can we do the same thing we did above where we don't define the key in both again, but rather do either output `[]` or the else block?

vivek-koppuru (changes requested):

helm-charts/stable/prow-control-plane/templates/config-ConfigMap.yaml
```diff
@@ -94,12 +94,15 @@ data:
...
+        - {{ $repo.org }}/{{ $repo.name }}
+          {{- end }}
       merge_method:
-      {{-  range $repo := .Values.repositories }}
-        {{ $repo.org }}/{{ $repo.name }}: squash
-      {{- end }}
+          {{-  range $repo := .Values.repositories }}
+          {{ $repo.org }}/{{ $repo.name }}: squash
+          {{- end }}
+          {{- end }}
```
micahhausler, line 107: What does this `end` correspond to?

abhinavmpandey08: The `end` on line 106 corresponds to `range` on line 104
The `end` on line 107 corresponds to `else` on line 99

micahhausler: oh great, can you adjust the indentation on the outermost if/else/end for clarity?

abhinavmpandey08: Yes, I will do that

micahhausler (approved):
/lgtm

vivek-koppuru (approved):

PR #101, opened 2021-02-04
Add support for different release strings

TerryHowe:
Add support for different formats of "release" by changing it to a string.

### DIFF
--- a/eks-distro-base/Makefile
+++ b/eks-distro-base/Makefile
@@ -10,7 +10,7 @@ IMAGE_TAG?=$(shell date "+%F-%s")
 IMAGE?=$(IMAGE_REPO)/$(IMAGE_NAME):$(IMAGE_TAG)
 
 ifeq ($(DEVELOPMENT),true)
-	BASE_IMAGE=amazonlinux:2
+	# BASE_IMAGE=amazonlinux:2
 endif
 
 ifeq ("$(REPO_OWNER)","")
@@ -36,16 +36,12 @@ local-images:
 .PHONY: images
 images:
 	# Sleeping until buildkit daemon is running on the other container
-	sleep 10
-	buildctl \
-		build \
-		--frontend dockerfile.v0 \
-		--opt platform=linux/amd64,linux/arm64 \
-		--opt build-arg:BASE_IMAGE=$(BASE_IMAGE) \
-		--local dockerfile=./ \
-		--local context=. \
-		--output type=image,oci-mediatypes=true,name=$(IMAGE),push=true
-	./update_base_image.sh $(IMAGE_TAG)
+	kubectl build --registry-secret ecr-public \
+		--platform=linux/amd64,linux/arm64 \
+		--build-arg BASE_IMAGE=$(BASE_IMAGE) \
+		--file Dockerfile \
+		--output type=image,oci-mediatypes=true,name=$(IMAGE),push=true .
+	# ./update_base_image.sh $(IMAGE_TAG)
 
 .PHONY: docker-public-login
 docker-public-login:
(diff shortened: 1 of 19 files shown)


eks-distro-base/Makefile
```diff
@@ -36,16 +36,12 @@ local-images:
...
-	buildctl \
-		build \
-		--frontend dockerfile.v0 \
-		--opt platform=linux/amd64,linux/arm64 \
-		--opt build-arg:BASE_IMAGE=$(BASE_IMAGE) \
-		--local dockerfile=./ \
-		--local context=. \
-		--output type=image,oci-mediatypes=true,name=$(IMAGE),push=true
-	./update_base_image.sh $(IMAGE_TAG)
+	kubectl build --registry-secret ecr-public \
```
abhay-krishna, line 39:
```suggestion
	buildctl build --registry-secret ecr-public \
```
Did we want to vend this image? Are we thinking in terms of how we vend the kube-proxy-base and go-runner base images?

release/api/v1alpha1/release.go
```diff
@@ -26,7 +26,7 @@ type ReleaseSpec struct {
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:Minimum=1
 	// Monotonically increasing release number
-	Number int `json:"number,omitempty"`
+	Release string `json:"number,omitempty"`
```
abhay-krishna, line 29:
```suggestion
	Release string `json:"release,omitempty"`
```

release/config/crds/distro.eks.amazonaws.com_releases.yaml
```diff
@@ -69,10 +55,10 @@ spec:
               type: string
             channel:
               type: string
-            number:
-              description: Monotonically increasing release number
+            release:
+              description: String indicating release
               minimum: 1
```
abhay-krishna, line 60: `minimum` field can be removed if we're moving to string.

eks-distro-base/Makefile
```diff
@@ -10,7 +10,7 @@ IMAGE_TAG?=$(shell date "+%F-%s")
 IMAGE?=$(IMAGE_REPO)/$(IMAGE_NAME):$(IMAGE_TAG)
 
 ifeq ($(DEVELOPMENT),true)
-	BASE_IMAGE=amazonlinux:2
+	# BASE_IMAGE=amazonlinux:2
```
vivek-koppuru, line 13: Why is this commented out?

eks-distro-base/Makefile
vivek-koppuru: Yea what are we trying to change here? I don't think kubectl supports this command so might be a mistake?

release/api/v1alpha1/release.go
vivek-koppuru: But would our release ever be something else other than a number? We shouldn't have to go with the CRD approach for postsubmits right? It is only to inform customers of the locations of the artifacts, but we would already have that configured in our scripts.

release/config/crds/distro.eks.amazonaws.com_releasechannels.yaml
```diff
@@ -1,17 +1,3 @@
-# Copyright 2020 Amazon.com Inc. or its affiliates. All Rights Reserved.
```
vivek-koppuru, line 1: Any reason we are removing the headers all together?

eks-distro-base/Makefile
TerryHowe: Yeh, sorry, I forgot to publish my comments. This is an unrelated change because I was rate limited by docker I think.

eks-distro-base/Makefile
TerryHowe: Sorry again, unrelated change because buildkit locally seems to have problems sometimes doing `yum update` getting mirrors. The kubectl build is an add-on.

TerryHowe:
I think we need to generate a numerical release number and not use SHA1

PR #107, opened 2021-02-24
Import upstream grafana chart and publish to S3 for PROW

mrajashree:
Currently PROW clusters use managed Grafana. This prevents us from using the declarative approach of adding dashboards for Grafana. So we are switching to using the upstream helm chart to install grafana and configure the dashboards.

### DIFF
--- a/helm-charts/scripts/install-toolchain.sh
+++ b/helm-charts/scripts/install-toolchain.sh
@@ -24,11 +24,13 @@ TMP_DIR="${BUILD_DIR}/tmp"
 mkdir -p "${TOOLS_DIR}"
 
 HELM_VERSION="v3.4.1"
-KUBECTL_VERSION=v1.18.9
+KUBECTL_VERSION=v1.19.6
 KIND_VERSION=v0.5.1
+RELEASE_BRANCH=1-19
+RELEASE=1
 
 ## Install kubectl
-curl -sSL "https://distro.eks.amazonaws.com/kubernetes-1-18/releases/1/artifacts/kubernetes/${KUBECTL_VERSION}/bin/linux/amd64/kubectl" -o "${TOOLS_DIR}/kubectl"
+curl -sSL "https://distro.eks.amazonaws.com/kubernetes-${RELEASE_BRANCH}/releases/${RELEASE}/artifacts/kubernetes/${KUBECTL_VERSION}/bin/linux/amd64/kubectl" -o "${TOOLS_DIR}/kubectl"
 chmod +x "${TOOLS_DIR}/kubectl"
 
 ## Install kubeval
--- a/projects/grafana/helm-charts/Makefile
+++ b/projects/grafana/helm-charts/Makefile
@@ -1,0 +1,41 @@
+GRAFANA_CHART_REPO?=helm-charts
+GRAFANA_CHART_CLONE_URL?=https://github.com/grafana/$(GRAFANA_CHART_REPO).git
+GRAFANA_CHART_GIT_TAG?=grafana-6.4.4
+
+ifeq ("$(GRAFANA_CHART_REPO)","")
+	$(error No repository name for Grafana chart was provided.)
+endif
+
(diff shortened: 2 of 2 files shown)


helm-charts/scripts/install-toolchain.sh
```diff
@@ -26,9 +26,10 @@ mkdir -p "${TOOLS_DIR}"
 HELM_VERSION="v3.4.1"
 KUBECTL_VERSION=v1.18.9
 KIND_VERSION=v0.5.1
+EKSD_VERSION=kubernetes-1-18
```
TerryHowe, line 29: I think the standard from the eks-d world would be to use RELEASE_BRANCH=1-18

abhay-krishna: I was gonna say the same

mrajashree: will change it 👍 
should we also switch to 1-19?

abhay-krishna: Yes I think we should be fore-runners in adopting our 1-19 release artifacts :wink:

mrajashree: changed it and also added one for RELEASE

projects/grafana/helm-charts/Makefile
```diff
@@ -0,0 +1,30 @@
+GRAFANA_CHART_REPO=helm-charts
+GRAFANA_CHART_CLONE_URL=https://github.com/grafana/$(GRAFANA_CHART_REPO).git
+GRAFANA_CHART_GIT_TAG=grafana-6.4.4
```
abhay-krishna, line 3: Can you add conditions to check these 3 values are not empty, like we do [here](https://github.com/aws/eks-distro-build-tooling/blob/main/projects/prometheus-community/helm-charts/Makefile#L10-L12)?

helm-charts/scripts/install-toolchain.sh
```diff
@@ -24,11 +24,13 @@ TMP_DIR="${BUILD_DIR}/tmp"
 mkdir -p "${TOOLS_DIR}"
 
 HELM_VERSION="v3.4.1"
-KUBECTL_VERSION=v1.18.9
+KUBECTL_VERSION=v1.19.6
 KIND_VERSION=v0.5.1
+RELEASE_BRANCH=kubernetes-1-19
```
TerryHowe, line 29:
```suggestion
RELEASE_BRANCH=1-19
```

helm-charts/scripts/install-toolchain.sh
```diff
@@ -24,11 +24,13 @@ TMP_DIR="${BUILD_DIR}/tmp"
...
+RELEASE=1
 
 ## Install kubectl
-curl -sSL "https://distro.eks.amazonaws.com/kubernetes-1-18/releases/1/artifacts/kubernetes/${KUBECTL_VERSION}/bin/linux/amd64/kubectl" -o "${TOOLS_DIR}/kubectl"
+curl -sSL "https://distro.eks.amazonaws.com/${RELEASE_BRANCH}/releases/${RELEASE}/artifacts/kubernetes/${KUBECTL_VERSION}/bin/linux/amd64/kubectl" -o "${TOOLS_DIR}/kubectl"
```
TerryHowe, line 33:
```suggestion
curl -sSL "https://distro.eks.amazonaws.com/${RELEASE_BRANCH}/releases/kubernetes-${RELEASE}/artifacts/kubernetes/${KUBECTL_VERSION}/bin/linux/amd64/kubectl" -o "${TOOLS_DIR}/kubectl"
```

projects/grafana/helm-charts/Makefile
mrajashree: sure, but why do we need to check it if we're setting it?

abhay-krishna (approved):
/lgtm

PR #109, opened 2021-02-28
Install bazel in builder-base image

abhay-krishna:
Some projects require Bazel for their builds, so we can bake it into our builder-base image.


TerryHowe (approved):
/lgtm

PR #110, opened 2021-03-01
Rename crds and create release channels

TerryHowe:
Make these names consistent with what is uploaded. Any new versions should be named with versions in them like this.


release/Makefile
```diff
@@ -24,13 +24,13 @@ run: generate fmt vet manifests
 
 # Install CRDs into a cluster
 install: manifests
-	kubectl apply -f ./config/crds/distro.eks.amazonaws.com_releasechannels.yaml
-	kubectl apply -f ./config/crds/distro.eks.amazonaws.com_releases.yaml
+	kubectl apply -f ./config/crds/releasechannels.distro.eks.amazonaws.com-v1alpha1.yaml
```
abhay-krishna, line 27: We would also need to rename the actual files referenced here right?

TerryHowe:
Waiting on https://github.com/aws/eks-distro-build-tooling/pull/107 I guess

mrajashree:
@TerryHowe #107 is merged now

abhay-krishna (approved):
/lgtm

PR #111, opened 2021-03-02
Increase prowjob age so that jobs stay longer on deck

abhay-krishna:
Make the prowjob CRD's more persistent by increasing their age. This will enable them to be displayed on deck for longer (1 month as opposed to 1 week by default), so we can have badges for infrequent jobs like release-postsubmits.

### DIFF
--- a/helm-charts/stable/prow-control-plane/templates/config-ConfigMap.yaml
+++ b/helm-charts/stable/prow-control-plane/templates/config-ConfigMap.yaml
@@ -24,6 +24,8 @@ data:
     sinker:
       # defaults to 1 hour
       resync_period: 2m
+      # approximately equal to 1 month, defaults to 1 week
+      max_prowjob_age: 730h
       # defaults to 1 day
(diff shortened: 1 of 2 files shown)


helm-charts/stable/prow-control-plane/templates/config-ConfigMap.yaml
```diff
@@ -24,6 +24,8 @@ data:
     sinker:
       # defaults to 1 hour
       resync_period: 2m
+      # approximately equal to 1 month, defaults to 1 week
+      max_prowjob_age: 730h
```
TerryHowe, line 28: I think the length of time between releases might be greater than a month, so maybe we shouldn't bother?

abhay-krishna: Yeah I was thinking the same about our release cadence, just set it to 1 month roughly but probably won't need to do it after the updated badge PR is merged.

PR #115, opened 2021-03-16
add generate-attribution script for generating ATTRIBUTION.txt files

jaxesn:
** Please review by commit **

This adds the generate-attribution script I wrote last year for the eks-d release to generate the ATTRIBUTION.txt files for each of our dependencies projects.  I have broken this up into 3 commits for easier review.

- adding generate script
- adding generate script + go-licenses to the builder-base
- adding test cases for generate script based on projects from eks-d.  These were added to make it easier if we ever need to make changes later and/or want to rewrite this in golang so its easier for others to maintain.

I have another PR coming on eks-distro to get this in use so that our ATTRIBUTION.txt files can be updated automatically when pushing new versions.


builder-base/Dockerfile
```diff
@@ -15,13 +15,37 @@
...
+ENV GOPATH /go
+ENV PATH="/go/bin/:$PATH"
+
+COPY *-checksum /
 COPY install.sh /opt/install.sh
+
 RUN bash /opt/install.sh
 
-ENV GOPATH /go
+# TODO: break the above install.sh script up into what is needed
```
jaxesn, line 26: Talking with @micahhausler  the original purpose of using the install.sh script was because the pod building the builder-base needed the same deps as builder-base itself.  This makes sense, however, at this point there is way more in the install.sh than is needed to build the builder-base so the original intent is not clear unless it is explained.  In addition to that, adding to the builder-base is painful because using the install script method does not allow docker to cache layers.

I propose doing the following for everything that is not required to build the builder-base to get the best of both of these worlds.  The install.sh script can still be used to bootstrap the pod and the builder-base, but everything else can be individual layers making rebuilds locally when making modifications less painful.

builder-base/Makefile
```diff
@@ -15,8 +15,16 @@ ifeq ($(DEVELOPMENT),true)
...
+copy-generate-attribution:
+	cp -rf ../generate-attribution ./
+
+.PHONY: remove-generate-attribution
+remove-generate-attribution:
+	rm -rf ./generate-attribution
+
 .PHONY: local-images
-local-images:
+local-images: copy-generate-attribution
```
vivek-koppuru, line 27: Can we just make these as part of the `build` and `release` targets so that we can have something like

```
build: copy-generate-attribution local-images remove-generate-attribution
```

builder-base/install.sh
```diff
@@ -70,6 +84,27 @@ sha256sum -c $BASE_DIR/buildkit-checksum
 tar -C /usr -xzf buildkit-$BUILDKIT_VERSION.linux-amd64.tar.gz
 rm -rf buildkit-$BUILDKIT_VERSION.linux-amd64.tar.gz
 
+if [[ "$CI" == "true" ]]; then
```
vivek-koppuru, line 87: When the script is run from inside a Dockerfile, this is false because we don't pass in that environment variable to the build right?

builder-base/install.sh
```diff
@@ -62,6 +64,18 @@ tar -C /usr/local -xzf go${GOLANG_VERSION}.linux-amd64.tar.gz
 rm go${GOLANG_VERSION}.linux-amd64.tar.gz
 mv /usr/local/go/bin/* /usr/bin/
 
+# install amazon-ecr-credential-helper
+# We are installing a specific commit of this because we need the sts regional endpoint changes in the aws sdk
+# to avoid hitting the global sts endpoint.
+# Commit: https://github.com/awslabs/amazon-ecr-credential-helper/commit/a004738dbac968cb287b47ae8ca39fd3b451e547
```
vivek-koppuru, line 70: Can we update this comment to mention tagged release instead because we changed it below?

builder-base/install.sh
jaxesn: Correct, this is the same conditional the script is using above to set the BASEDIR var only in the case thats its not running in the docker container.

builder-base/install.sh
jaxesn: There is actually a v0.5.0 release in amazonlinux now, switching to install that via yum.

```
bash-4.3# yum info amazon-ecr-credential-helper  
Loaded plugins: ovl, priorities
Available Packages
Name        : amazon-ecr-credential-helper
Arch        : x86_64
Version     : 0.5.0
Release     : 1.amzn2
Size        : 2.3 M
Repo        : amzn2extra-docker/2/x86_64
Summary     : Amazon ECR Docker Credential Helper
License     : Apache 2.0
Description : The Amazon ECR Docker Credential Helper is a credential helper for the Docker
            : daemon that makes it easier to use Amazon Elastic Container Registry.

```

https://github.com/awslabs/amazon-ecr-credential-helper/releases/tag/amzn2%2F0.5.0-1

builder-base/Makefile
jaxesn: good idea!

TerryHowe (approved):
/lgtm

PR #117, opened 2021-03-16
Add script to run buildctl in daemonless mode

abhay-krishna:
Adding script to builder-base image to perform image builds in CodeBuild in daemonless mode. This script was largely inspired from https://github.com/moby/buildkit/blob/master/examples/buildctl-daemonless/buildctl-daemonless.sh, and slightly modified to suit our usecase.

### DIFF
--- a/builder-base/buildkit.sh
+++ b/builder-base/buildkit.sh
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+# Copyright 2020 Amazon.com Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Script to perform buildkit builds in daemonless mode
+
+set -e
+set -o pipefail
+set -x
+
+tmp=$(mktemp -d /tmp/buildctl-daemonless.XXXXXX)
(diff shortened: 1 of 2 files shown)


jaxesn (commented):
This seems reasonable to me.  Have you tested it in the codebuild env?

abhay-krishna:
> This seems reasonable to me. Have you tested it in the codebuild env?

I had to set privileged flag to true in CodeBuild, since it was complaining about not being able to perform some mounting operations otherwise.
```bash
error: failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to read dockerfile: failed to mount /tmp/buildkit-mount447148898: [{Type:bind Source:/var/lib/buildkit/runc-native/snapshots/snapshots/2 Options:[rbind ro]}]: operation not permitted
```

builder-base/buildkit.sh
```diff
@@ -0,0 +1,49 @@
...
+
+set -e
+set -o pipefail
+set -x
+
+tmp=$(mktemp -d /tmp/buildctl-daemonless.XXXXXX)
+trap "kill \$(cat $tmp/pid); rm -rf $tmp" EXIT
+
+start::buildkitd() {
+    buildkitd --addr=unix:///run/buildkit/buildkitd.sock --oci-worker-platform=linux/amd64 --oci-worker-platform=linux/arm64 >$tmp/log 2>&1 &
```
vivek-koppuru, line 26: Are you missing the flag defined here? https://github.com/aws/eks-distro-prow-jobs/blob/main/scripts/buildkitd-entrypoint.sh#L21

abhay-krishna: We can use that flag only if the command is invoked with Rootlesskit. Within the security context of a single container, it is complicated to allow only the buildkit daemon to run as a rootless user with the rest of the build operations as root. Had a discussion with Vignesh and Micah about this, and we reached a consensus that we will not be using rootless here. Also these builds are ephemeral and isolated, and the script has logic to kill the daemon even if the build fails, so it should be okay to run as root.

jaxesn (approved):
/lgtm

PR #119, opened 2021-03-17
Prevent heredoc from expanding Bash exit code variable

abhay-krishna:
The Dockerfile generated was being hardcoded with an exit code of 0 since the Heredoc did the variable expansion of `$?` instead of keeping it as such.


TerryHowe (approved):
/lgtm

jaxesn (approved):
/lgtm

PR #125, opened 2021-03-29
Adding image-builder dependencies to builder-base

vignesh-goutham:
Adding packer, ansible, goss and pywinrm to builder-base.

These dependencies are used by kubernetes-sigs/image-builder to create capi images.

We will be removing older version of packer and installing a more stable and newer version with this change.

### DIFF
--- a/builder-base/install.sh
+++ b/builder-base/install.sh
@@ -103,29 +103,34 @@ yum install -y \
     which
 
 # Install image-builder build dependencies
+# Post upgrade, pip3 got renamed to pip and moved locations. It works completely with python3
+# Symlinking pip3 to pip, to have pip3 commands work successfully
 pip3 install -U pip setuptools
+ln -s /usr/local/bin/pip /usr/bin/pip3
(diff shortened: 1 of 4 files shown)


builder-base/install.sh
```diff
@@ -102,13 +102,35 @@ yum install -y \
     vim \
     which
 
-ROOTLESSKIT_VERSION="${ROOTLESSKIT_VERSION:-v0.13.2}"
+# Install image-builder build dependencies
+# Post upgrade, pip3 got renamed to pip and moved locations. It works completely with python3
+# Symlinking pip3 to pip, to have pip3 commands work successfully
+pip3 install -U pip setuptools
+ln -s /usr/local/bin/pip /usr/bin/pip3
```
TerryHowe, line 109: Looking at these paths and I'm getting nervous, perhaps this is not the best approach. The normal Pythonic way of doing things is with virtual environments. Whatever version of Python or pip you have to start, this should work.

```
pip install virtualenv
virtualenv venv
source venv/bin/activate
pip install pip # etc
```

PR #128, opened 2021-04-01
Untaring goss onto imagebuilder home paths

vignesh-goutham:
Image-builder requires goss to be present under the imagebuilder user home path.

This change untars the goss plugin onto the right path.


abhay-krishna (approved):
/lgtm

PR #130, opened 2021-04-07
for deps that are replaced in the go.mod pull that replace version

jaxesn:
While working on generating the [attribution](https://github.com/aws/modelrocket-build-tooling/blob/main/projects/kubernetes-sigs/cri-tools/ATTRIBUTION.txt#L44) for critools, we noticed that certain dependencies are replaced in [go.mod](https://github.com/kubernetes-sigs/cri-tools/blob/master/go.mod#L33) file.  In cases like this the most correct version is the one referenced in the replace lines.  


TerryHowe (approved):
/lgtm

PR #133, opened 2021-04-16
Update golang versions

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*

Updating the go versions and also the default go for our builder base. This should not break existing dependencies because the ones that are tied to specific versions have overrides, and we should always be updating our default go version in the builder base. This is also needed to consume the embed package for the CLI.

### DIFF
--- a/builder-base/install.sh
+++ b/builder-base/install.sh
@@ -54,7 +54,7 @@ aws --version
 rm awscli-exe-linux-x86_64.zip
 rm -rf /aws
 
-GOLANG_VERSION="${GOLANG_VERSION:-1.15.10}"
+GOLANG_VERSION="${GOLANG_VERSION:-1.16.3}"
 wget \
     --progress dot:giga \
     --max-redirect=1 \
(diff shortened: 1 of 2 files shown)


builder-base/install.sh
```diff
@@ -178,7 +178,8 @@ setupgo() {
 
 setupgo "${GOLANG113_VERSION:-1.13.15}"
 setupgo "${GOLANG114_VERSION:-1.14.15}"
-setupgo "${GOLANG115_VERSION:-1.15.10}"
+setupgo "${GOLANG115_VERSION:-1.15.11}"
+setupgo "${GOLANG115_VERSION:-1.16.3}"
```
pokearu, line 182: I think this should be 
`${GOLANG116_VERSION:-1.16.3}` ?

vivek-koppuru: Good catch!

jaxesn (approved):
The next run of the attribution job will trigger an update PR in eks-distro bumping to 1.15.11. Not a big deal, just making it known.

/lgtm

PR #134, opened 2021-04-16
Add golangci-lint to builder-base

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/builder-base/install.sh
+++ b/builder-base/install.sh
@@ -65,6 +65,9 @@ tar -C /usr/local -xzf go${GOLANG_VERSION}.linux-amd64.tar.gz
 rm go${GOLANG_VERSION}.linux-amd64.tar.gz
 mv /usr/local/go/bin/* /usr/bin/
 
+# binary will be $(go env GOPATH)/bin/golangci-lint
+curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin ${GOLANGCILINT_VERSION:-v1.39.0}
+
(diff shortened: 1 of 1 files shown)


builder-base/install.sh
```diff
@@ -65,6 +65,9 @@ tar -C /usr/local -xzf go${GOLANG_VERSION}.linux-amd64.tar.gz
 rm go${GOLANG_VERSION}.linux-amd64.tar.gz
 mv /usr/local/go/bin/* /usr/bin/
 
+# binary will be $(go env GOPATH)/bin/golangci-lint
+curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin ${GOLANGCILINT_VERSION:-v1.39.0}
```
g-gaston, line 69: I'm ok merging this (better have something than nothing) but I think it is a bit dangerous
Each version of golanci-lint introduces new linters of updates the ones already present
Hence, depending on when this image is built, the linters we will be running might be different. Which makes the build non deterministic. If the image is updated, we might be introducing new linter rules, and the PR jobs will start failing all of the sudden
Also, updating the version in our go.mod won't update the version here. So we will be running something difference in our machines than here

Would it be possible to make this process (getting the golangci-lint) part of our `make` process, using the version specified in out go.mod?

vivek-koppuru: Good point, I honestly wish we had something a bit more native for linting, like running something like go fmt. I agree with the Makefile approach and will discard this PR.

pokearu: But since we have a .golangci.yml would it not always enable only what we specify there?
But I guess the version can be a problem.

PR #137, opened 2021-04-19
Use public ECR eks-distro-base as base image

abhay-krishna:
Now that we have eks-distro-base image in public ECR, we can use them as base images for building EKS-D.


projects/gomods/athens/Makefile
```diff
@@ -6,7 +6,7 @@ AWS_ACCOUNT_ID?=$(shell aws sts get-caller-identity --query Account --output tex
 AWS_REGION?=us-west-2
 
 MAKE_ROOT=$(shell cd "$(shell dirname "${BASH_SOURCE[0]}")" && pwd -P)
-BASE_REPO?=$(AWS_ACCOUNT_ID).dkr.ecr.$(AWS_REGION).amazonaws.com
+BASE_REPO?="public.ecr.aws/h1r8a7l5"
```
pokearu, line 9: This looks good. But is there a way to pass this in instead of pointing to the public by default?

PR #138, opened 2021-04-22
Modify script to automate builder-base image tag update in tag file

abhay-krishna:
We are tracking builder-base image in a separate file. The automated PR should update the tag in that file too when a new image is created.


TerryHowe (approved):
/lgtm

PR #139, opened 2021-04-24
Set up base image pushes to eks-distro-build-tooling ECR public registry

abhay-krishna:
This setup script makes sure we use the proper Role to push to the `eks-distro-build-tooling` repo where we have the base image repos.

/hold


jaxesn (approved):
/lgtm

PR #140, opened 2021-04-27
attribution: better handling of relative path replace paths

jaxesn:
This came up when adding attribution generation for capd since it pulls in capi deps from higher up in the repo path (../../)


abhay-krishna (approved):
/lgtm

PR #142, opened 2021-04-27
Add reference to public ECR base image repos in respective READMEs

abhay-krishna:
This will also trigger the base image builds and pushes.


TerryHowe (approved):
/lgtm

PR #145, opened 2021-04-28
Perform ecr-setup step in presubmit

abhay-krishna:
Ensuring that env variables are set as expected.


pokearu (approved):
/lgtm

PR #147, opened 2021-04-28
Remove unnecessary exports from script

abhay-krishna:
We set these as env vars in the postsubmits so we don't need them here.


pokearu (approved):
/lgtm

PR #149, opened 2021-04-29
Check if ROLE_ARNs are set properly in script

abhay-krishna:
Validate proper configuration of role_arn's


pokearu (approved):
/lgtm

PR #153, opened 2021-04-29
Remove DEVELOPMENT var from build-tooling projet Makefiles

abhay-krishna:
We don't need this variable anymore, since we are consuming public ECR images by default.


pokearu (approved):
/lgtm

PR #155, opened 2021-04-30
Fix regex for proper sed replacement

abhay-krishna:
Since the image repo name changed, we need to fix the search regex in this script.


jaxesn (approved):
/lgtm

PR #157, opened 2021-05-01
install gettext on builder base for envsubst

jaxesn:
For generating config files on the fly based one eks-distro release yaml, Id like to use envsubst which is provided by gettext.


abhay-krishna (approved):
/lgtm

PR #162, opened 2021-05-11
Remove unnecessary args when invoking script in periodic

abhay-krishna:
Since we've hardcoded the image name and image repo, we don't need to pass them in as arguments to the script.


pokearu (approved):
/lgtm

PR #165, opened 2021-05-13
Add script to check if buildkit daemon is running

abhay-krishna:
Run a script to check if buildkit daemon is running before invoking buildkit client


builder-base/Makefile
```diff
@@ -19,9 +25,7 @@ remove-generate-attribution:
 	rm -rf ./generate-attribution
 
 .PHONY: local-images
-local-images:
-	# Sleeping until buildkit daemon is running on the other container
-	sleep 10
+local-images: buildkit-check
```
jaxesn, line 28: Instead of doing this here, could we do it in the prow job instead?  Doing it here could complicate the local setup.

pokearu (approved):
/lgtm

PR #169, opened 2021-05-18
Add buildkit check for postsubmit and periodic make targets

abhay-krishna:
We need the buildkit check for postsubmits and periodics since we call `buildctl` here as well.


pokearu (approved):
lgtm

pokearu (approved):
/lgtm

PR #171, opened 2021-05-29
Configure ECR Public push setup script and make target to run in same shell

abhay-krishna:
The env vars set by the script are not available in the same shell as the one in which the make target is invoked.


pokearu (approved):
/lgtm

PR #174, opened 2021-06-25
Use Bash shell in Makefile to source script

abhay-krishna:
Makefile by default uses `sh` shell, which doesn't implement source. Hence we are changing it to use `bash` shell so that the periodic can source the setup script.


pokearu (approved):
/lgtm

PR #179, opened 2021-07-06
Avoid re-running check-update script in periodic

abhay-krishna:
Fixing the Makefile to run the check-update script only once during the periodic.
/hold


pokearu (approved):
/lgtm

PR #183, opened 2021-07-23
Add auth header with github token for reading licenses

abhay-krishna:
Passing auth token in header so that rate limiting doesn't occur.

### DIFF
--- a/generate-attribution/generate-attribution-file.js
+++ b/generate-attribution/generate-attribution-file.js
@@ -153,7 +153,8 @@ async function addGoLicense(dependencies) {
 async function readLicenseFromUpstream(upstreamUrl) {
     return new Promise((resolve, reject) => {
         let finalDoc = '';
-        const req = https.get(upstreamUrl, res => {
+        const options = await generateAuthorizationHeader()
+        const req = https.get(upstreamUrl, options, res => {
             res.on('data', d => {
                 finalDoc += d;
             })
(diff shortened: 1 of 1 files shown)


generate-attribution/generate-attribution-file.js
```diff
@@ -338,6 +340,15 @@ async function populateVersionAndModuleFromDep(dependencies) {
     return finalDeps;
 }
 
+function generateAuthorizationHeader() {
+    const githubToken = await fsPromises.readFile("/secrets/github-secrets/token", 'utf8');
```
jaxesn, line 344: Probably not the most elegant but just add a .stat call before hand to make sure the file exists. If it doesn't, just return an empty object. I think ideally wed pass this in as a cli arg or use an env var, but this is totally fine

generate-attribution/generate-attribution-file.js
```diff
@@ -338,6 +340,23 @@ async function populateVersionAndModuleFromDep(dependencies) {
     return finalDeps;
 }
 
+function generateAuthorizationHeader() {
+    const githubTokenFile = "/secrets/github-secrets/token";
+    fs.stat(githubTokenFile, (error, stats) => {
```
jaxesn, line 345: Sorry this is the pattern to follow, in the catch block return empty object, that means file doesn't exist

https://github.com/aws/eks-distro-build-tooling/blob/ac2191b893bfde5628e758007c7fda5306cd7e36/generate-attribution/generate-attribution-file.js#L220

abhay-krishna: Thanks @jaxesn, updated the PR accordingly.

generate-attribution/generate-attribution-file.js
```diff
@@ -338,6 +340,21 @@ async function populateVersionAndModuleFromDep(dependencies) {
     return finalDeps;
 }
 
+async function generateAuthorizationHeader() {
+    const githubTokenFile = "/secrets/github-secrets/token";
+    try {
+        await fsPromises.access(githubTokenFile);
+        const githubToken = await fsPromises.readFile(possiblePath, 'utf8');
```
jaxesn, line 347:
```suggestion
        const githubToken = await fsPromises.readFile(githubTokenFile, 'utf8');
```

jaxesn (approved):
/lgtm

PR #185, opened 2021-07-23
Call async function outside return statement

abhay-krishna:
Waiting for an async method inside a Promise closure is kind of an anti-pattern.


TerryHowe (approved):
/lgtm

PR #190, opened 2021-07-26
Create pull-requests conditionally in periodic

abhay-krishna:
The create-pr target is called by default in the Prowjobs, and while we want to call the PR script for presubmits (dry-run) and postsubmit (always create PR due to deterministic image push), we need a condition to check if images were pushed in periodic, to decide if a PR is to be created.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ release/cover.out
 .idea
 index.yaml
 eks-distro-base/check-update/
+eks-distro-base/*-pushed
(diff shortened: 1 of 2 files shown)


micahhausler (commented):
Can you add the `-pushed` file to the `.gitignore`?

abhay-krishna:
> Can you add the `-pushed` file to the `.gitignore`?

Done. Great catch!

pokearu (approved):
/lgtm

PR #192, opened 2021-07-29
minimal image: add nonroot variant and clean up static files

jaxesn:
- Instead of using nsswitch from al2, use a more generic similar to distroless
- add passwd/group files to define root and nonroot users
- add os-release file similar to distroless 

### DIFF
--- a/eks-distro-base/Dockerfile.minimal-nonroot
+++ b/eks-distro-base/Dockerfile.minimal-nonroot
@@ -1,0 +1,18 @@
+# Copyright 2020 Amazon.com Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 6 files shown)


eks-distro-base/Dockerfile.minimal-nonroot
```diff
@@ -0,0 +1,18 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+ARG BASE_IMAGE
+FROM ${BASE_IMAGE} 
+
```
micahhausler, line 17: Should we `chown 65532:65532 /home/nonroot`?

jaxesn: good question.. let me look at what distroless does for that

jaxesn: couldnt find where in distroless they set this, but Im thinking you are right so added it to the base dockerfile.

micahhausler (approved):
/lgtm
/approve

PR #201, opened 2021-08-10
minimal: better usage of rpmdb to verify dependencies

jaxesn:
The main purpose of this PR is to make better use of rpm and yumdownloader to avoid having to hardcode excluded dependencies across different variants.  This came about when I started looking at potentially producing a variant to support csi drivers, which require a number of packages (xfs-progs, linux-utils, etc) and wanted to get a better system in place to catching any missing dep and make it more clear when we are excluding some specifically.  

Using rpm and yum to install the packages instead of extracting ensures pre/post install scripts are running giving us better confidence we have the right content in these images.  This does require a "[hack](https://github.com/aws/eks-distro-build-tooling/pull/201/files#diff-156d44d349c8605b3441a4e143a50ebbd8e39a070cadadfaff02e7182bdf1c22R109)" to get around missing dependencies and avoiding installation of entire trees worth of deps we don't want.  `systemd` is one that we do not want in any image, to avoid installing it and its dep, it is manually downloaded and installed into the rpmdb and removed before final image is exported.  For all but the csi variant, bash is also not wanted in the final image.  Bash is however needed when installing packages so it is typically added and then removed before export.

Probably best to review by commit.

Changes by commit:
1. 
  - download base rpms and manually install, then rely on yum to install others
  - better cleanup of doc and man folders to avoid deleting the various license files (LICENSE, COPYING, COPYRIGHT, NOTICE)
  - introduce a glibc variant in between base and iptables. eventually well probably need to add openssl to this to support building fips compliant go binaries, but skipping that for now.  glibc will also be needed when adding the csi variant
  - add iptables-nft to fix https://github.com/aws/eks-distro-build-tooling/issues/202.  I am not adding the iptables-wrapper script that kubernetes/release provides since that seems like something we should add in the eks-d/eks pipelines, which we do already today.  NOTE, this does not really solve the issue because the [iptables-wrapper](https://github.com/kubernetes/release/blob/master/images/build/debian-iptables/buster/iptables-wrapper) is a bash script which we are intentionally not including in our base images.  Adding nft + update-alternatives may still make sense so people could build a new image from this base and change the alternative to nft.  The only option would be for us to implement the wrapper script in golang, this probably wouldn't be too difficult.

2. introduce basic tests to exercise images, uses docker so intended to work only locally

3. add exporter image to export packages which make up the different images.  This will help us audit as images are auto updated

4. add csi variant to support https://github.com/aws/modelrocket-build-tooling/blob/main/projects/kubernetes-sigs/vsphere-csi-driver/docker/linux/csi-driver/Dockerfile

TODO
- [x]  add -glibc/-csi variant to CDK to create the new repo


eks-distro-base/Dockerfile.minimal
```diff
@@ -25,43 +28,52 @@ RUN yum upgrade -y && \
...
-RUN rpm --initdb --dbpath /newroot/var/lib/rpm && \
-    rpm -ivh --nodeps --justdb --dbpath /newroot/var/lib/rpm $DOWNLOAD_DIR/*.rpm && \
-    rpm --dbpath /newroot/var/lib/rpm -qa | sort
+COPY files/ $NEWROOT
 
 
 ################# BASE ########################
 FROM scratch as base
 ENV PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
-ENV SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt
```
jaxesn, line 64: this was the wrong path and this var isnt needed anyway.  golang will automatically scan a list of known paths which include where our bundle exists.

abhay-krishna (approved):
/lgtm

PR #205, opened 2021-08-21
only export if the image has been pushed

jaxesn:
Avoid building images just to export, only run export on images that built/pushed.


abhay-krishna (approved):
/lgtm

PR #218, opened 2021-09-15
Added nodeSelector, tolerations, affinity

kwangil-ha:


jaxesn:
Thanks for PR!  Could you give us more details about the change and why this is important to you?  We use this chart to support our build infra for EKS-D, its not really intended as a deliverable for use elsewhere.  There is no reason you can't thought, I just want to understand more about the use case.

PR #239, opened 2021-10-08
eks-distro-base: add latest tag

jaxesn:
*Issue #, if available:*

*Description of changes:*

We are starting to get use of these images outside of eks-distro repos.  As a convenience, introducing a latest tag.   We will still use tags across our repos to be more specific and reproducible.


abhay-krishna:
> I thought we were avoiding this because it is nondeterministic

We're gonna be using the tagged versions for all the base image stuff, this is just in addition to that.

PR #265, opened 2021-11-19
Add affinity spec to Helm chart values and templates

abhay-krishna:
This is to help us constrain what nodes these pods get deployed on.

**Note:** I chose to use `nodeAffinity` instead of `nodeSelector` since the former is a superclass of and provides a more expressive way to schedule nodes than the latter, and at some point, `nodeSelector` will be deprecated (although not until K8s launches v2 API).


jaxesn (approved):
/lgtm

PR #266, opened 2021-11-25
Add eks-anywhere repo to base image tag update workflow

abhay-krishna:
The EKS Anywhere cluster controller image uses the minimal eks-d base as the base image, so that tag file needs to be update accordingly when a new image is pushed.


jaxesn (approved):
/lgtm

PR #271, opened 2021-12-05
Add override plugin, remove needs-rebase plugin

abhay-krishna:
The `override` plugins allows PR authors to force job statuses on Github to pass. This is especially useful when adding new blocking presubmits (for example, adding presubmit for a new project) which causes the status reconciler to run the job on all open PRs and fail citing the path references by the job doesn't exist ([example](https://github.com/aws/eks-anywhere-build-tooling/pull/228)).

The `needs-rebase` plugin is an external plugin whose events are not handled by Prow's hook component but through an external endpoint that has been configured to handle GitHub webhooks. We don't have support for it at this point, so removing it.


TerryHowe (approved):
/lgtm

PR #273, opened 2021-12-09
Bumping golang versions and adding 1.17

pokearu:
*Description of changes:*
Adding Go 1.17 to builder base and bumping other golang versions.


builder-base/install.sh
```diff
@@ -50,7 +50,7 @@ yum install -y \
     unzip \
     wget
 
-GOLANG_VERSION="${GOLANG_VERSION:-1.16.7}"
+GOLANG_VERSION="${GOLANG_VERSION:-1.16.12}"
```
jaxesn, line 53: youll need to update the checksum files for this one

jaxesn (commented):
When this is merged and a new image is pushed, we should probably trigger the two attribution periodic jobs which will also build everything and open PRs to update any changed checksums.  If we dont, then builds in build-tooling may start "randomly" failing due to checksum mismatches.

builder-base/install.sh
pokearu: Got it thanks!

abhay-krishna (approved):
/lgtm

PR #276, opened 2021-12-15
distro-base: kind + haproxy + nginx new minimal variants

jaxesn:
*Issue #, if available:*

*Description of changes:*

Adding new variants to be used by kind and the eks-anywhere-test image.  Pulling these into here allows us to make sure the yum packages are all updated automatically and keeps all yum installs in the same place.

/hold

TODO:

- [ ] add new ecr repos (@abhay-krishna)


TerryHowe:
/lgtm
When you are ready. Looks good! I'm sure we'll find other uses for these.

jaxesn:
Merging, this will not kick off a postsubmit, instead i will trigger the periodic.

PR #280, opened 2021-12-17
Add exec permissions to yq binary

abhay-krishna:
YQ download was refactored to download the binary directly from the releases instead of from the tarball and unpacking it. Turns out the binary inside the tarball has exec perms on it,  but the standalone binary from release does not.


jaxesn (approved):
/lgtm

PR #295, opened 2022-02-08
Modify url for dev releases

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*

Currently the dev release manifests are pointing to the prod cdn, which is causing our builds from the EKS-A side to fail because the artifacts don't exist in the urls specified. This will read the bucket from the env var and if the dev-release flag is set, use the dev url instead.


abhay-krishna (approved):
/lgtm

PR #308, opened 2022-02-16
builder-base: add al2022 build

jaxesn:
*Issue #, if available:*

*Description of changes:*
This introduces an al2022 based builder-base image.  It also changes the base image from al2/2022 directly to the minimal kind image, which has most of the same stuff this image needs.  

Much like the eks-distro-base build, this adds an AL_TAG which is meant to be used to determine the underlying base image.  The al2 build continues to be tagged latest, as well as latest.2, and the 2022 build will be tagged with latest.2022.


abhay-krishna (approved):
/lgtm

PR #309, opened 2022-02-16
Add build steps for Prow deck image

abhay-krishna:
This PR adds build logic for the upstream Kubernetes test infra Deck image for Prow, with the change being the embedded style and script instead of fetching from a CDN.


projects/kubernetes/test-infra/update_helm_chart.sh
```diff
@@ -0,0 +1,71 @@
...
+set -e
+set -o pipefail
+set -x
+
+PATCH=patch
+if [[ "$(uname -s)" == "Darwin" ]]; then
+    PATCH=gpatch
+fi
+
+if ! command -v $PATCH > /dev/null; then
```
jaxesn, line 26: meh, i dont know about this.  We have deps all over that need to be installed and we dont typically try to install them.  Not saying we dont need to do better, but im not sure this is the best approach.  You decide tho if you want to keep it.

jaxesn (commented):
Im fine with this, small nit you decide.

jaxesn (approved):
/lgtm

PR #316, opened 2022-02-24
release: kube generate image tags like all other container images

jaxesn:
*Issue #, if available:*

*Description of changes:*

In https://github.com/aws/eks-distro/pull/753 I am looking to change the KUBE_GET_VERSION in the version file the release tooling uses to generate the final release yaml.  Image tags and binary versions will not match on the kubernetes components.  This change uses the same process to generate the image tags as all the other components in the release tooling. 

At this point the only thing the release tooling will be pulling from the version file is the git commit hash.

/hold


kschumy (approved):
/lgtm

PR #318, opened 2022-02-25
builder-base: set root user shell to bash instead of nologin

jaxesn:
*Issue #, if available:*

*Description of changes:*
We ran into an issue yesterday in codebuild where a job tried to switched to the root user and was not able to because the shell was improperly configured.


abhay-krishna (approved):
/lgtm

PR #322, opened 2022-03-01
distro-base: check in updates file for use in pr creation

jaxesn:
*Issue #, if available:*

*Description of changes:*
The last periodic job that update the distro base images did not check in the updates file which tracks the packages that were changed.  This meant that when the other PRs were opened, the message showed the previous updated packages instead of the latest.  This should fix that.

I took a simple approach here... the ADDITIONAL_GIT_ADD was only used in this one spot and the quotes around multiple were messing up the git add.  Instead of improving that for one case and since this script is already hardcoding cases for different repos, I took this approach.

### DIFF
--- a/eks-distro-base/Makefile
+++ b/eks-distro-base/Makefile
@@ -199,7 +199,7 @@ endif
 
 .PHONY: create-pr
 create-pr: $(CREATE_PR_TARGETS)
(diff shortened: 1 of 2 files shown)


abhay-krishna (approved):
/lgtm

PR #331, opened 2022-03-11
Change apiextensions to use v1 api

vivek-koppuru:
*Issue #, if available:*

*Description of changes:*
1.22 deprecates v1beta1 apiextensions: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22

Made changes by applying the crd on a 1.21 cluster and retrieving the v1 version of it so it would automatically convert it for me.


jaxesn:
Will this work on a 1.20 cluster?

vivek-koppuru:
> Will this work on a 1.20 cluster?

It works on any cluster after 1.16

PR #333, opened 2022-03-11
Update skopeo release

TerryHowe:


TerryHowe:
Tag is wrong in tags file looks like

PR #334, opened 2022-03-12
Fix 2022 base images

TerryHowe:
This is probably wrong, but closer. I assume I'm supposed to generate these, but I just got the latest tags I think are right.


jaxesn:
The periodic should have built these and updated. Let me check it out

PR #335, opened 2022-03-12
distro-base: fix update_packges and update image tag

jaxesn:
*Issue #, if available:*

*Description of changes:*
In the last periodic build a couple things happened.

- the update_packages were not being checked in so the PR messages in other repos were still wrong
- the update base image tag script was also updating al2022 when it should have only been updating the al2 ones. This is because the image tags are the same between the two.


TerryHowe (approved):
/lgtm

PR #339, opened 2022-03-21
Upgrading go default to 1.17.5

jonahjon:
Signed-off-by: jonahjon <jonahjones094@gmail.com>

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/builder-base/install.sh
+++ b/builder-base/install.sh
@@ -185,8 +185,6 @@ setupgo() {
     ln -s /root/sdk/go${version}/bin/gofmt ${GOPATH}/go${majorversion}/bin/gofmt
 }
 
-setupgo "${GOLANG117_VERSION:-1.17.5}"
-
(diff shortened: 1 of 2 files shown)


jaxesn:
Let's hold off on making go lang 1.17 the default. This should not be necessary since the image already has 1.17 in it

jonahjon:
> Let's hold off on making go lang 1.17 the default. This should not be necessary since the image already has 1.17 in it

I think the issue is with the e2e prow jobs. We might have to override the make all e2e commands from `go` > `go1.17.5` as an alternative to make things work for our bug bash.

jaxesn:
You should be able to follow the same process we do for all the other build where the go1.17 bin path is prepended to the PATH env var to force it.

https://github.com/aws/eks-anywhere/blob/main/Makefile#L24

We should avoid depending on the default installed version of go from builder base so we dont run into issues later when we do upgrade it.

PR #345, opened 2022-03-30
Fix Deck service ports mapping

abhay-krishna:
This was removing a preceding `-` that caused validations since it thought it was a map and not an array.


g-gaston (approved):
/lgtm

PR #346, opened 2022-03-30
Use proper git diff target in Helm version check

abhay-krishna:
Patch folder needs to be added to the diff targets, not replace them.


kschumy (approved):
/lgtm

TerryHowe (approved):
/lgtm

PR #348, opened 2022-04-01
builder-base: update 1.16 and 1.17, add 1.18 and update go-licenses

jaxesn:
*Issue #, if available:*

*Description of changes:*

This bumps to the latest 1.16 and 1.17 golang versions.  1.18 will not be included for builds who need to upgrade.  1.16 is still the default version, this is mostly due to some uncertainty around go-licenses and attribution generation. I *think* we could change this, but want more time to validate that.  Builds should not be relying on the default install anywhere.

go-licenses recently pushed a 1.0.0 tag so updating to it.

There was a bug in the shasum collection, at least on mac, which I fixed.


abhay-krishna (approved):
/lgtm

builder-base/install.sh
```diff
@@ -185,7 +185,8 @@ setupgo() {
     ln -s /root/sdk/go${version}/bin/gofmt ${GOPATH}/go${majorversion}/bin/gofmt
 }
 
-setupgo "${GOLANG117_VERSION:-1.17.5}"
+setupgo "${GOLANG117_VERSION:-1.17.8}"
+setupgo "${GOLANG117_VERSION:-1.18}"
```
RedbackThomson, line 189: Should this have been `GOLANG118_VERSION`?

PR #353, opened 2022-04-04
eks-distro-base: avoid relying on builtkit cache by using built builder

jaxesn:
*Issue #, if available:*

*Description of changes:*

We recently enabled gc on buildkit in prow and now in some of our builds the image is built twice, once for the actual image and once for the builder target.  This is because we run two buildkit builds to generate these images and before the second runs would be no-ops since they were using the buildkit cache.  This could probably be solved via tuning the gc settings, however, I have wanted for a while to take another pass at this makefile and set it up so that the different "stages" of each image build could be based off the pushed image instead of the cache.

This PR does a lot:

- Introduce a new var to image builders, `BUILT_BUILDER_IMAGE`, which is the pushed builder stage of the current image. The `base` and `export` stages rely on this instead of the `builder` stage to copy and export.  This changes mean that as long as the builder is pushed, its easy for us to recreate the other two stages by calling the target with the proper var.  You can see this in use with the new `packages-export-all-minimal-images` target
- refactor make file to avoid calling $(MAKE) directly and instead use target pre-reqs.  This also required changing the `if` blocks to be more `make-y` so that could properly be evaluated per target.
- the minimal image "dependency tree" is now "modeled" via target pre-reqs vs the older if block, this feels more extensionable
- The image targets more closely align with the Common.mk strategy in other repos, tho its not a direct 1:1.
- the `base` and `export` stages are now mostly standardized in the new Dockerfile.minimal-helpers file since they are mostly all the same and with the new `BUILT_BUILDER_IMAGE` strategy we do not need to be stages in the same file.
- `packages-export-all-minimal-images` was introduced to manually update the packages files for cases, which happens regularly, where the automated build does not check these in.


eks-distro-base/Dockerfile.minimal-base
```diff
@@ -60,13 +57,9 @@ COPY files/ $NEWROOT
 
 
 ################# BASE ########################
+FROM ${BUILT_BUILDER_IMAGE} as base-builder
 FROM ${BASE_IMAGE} as base
```
jaxesn, line 61: I actually wouldnt mind a better name for this stage. We currently have 3:

builder
base
export

I think builder and export are fairly explanatory and base kind of makes sense, however the word `base` is also used to identify the most base minimal image.  It might be a nice change if we had a better name for this stage to avoid the ambiguity?

abhay-krishna: How about something like `final` or `core`?

abhay-krishna (approved):
/lgtm

PR #361, opened 2022-04-10
distro-base: add arm64 build of symlinks

jaxesn:
*Issue #, if available:*

*Description of changes:*

The symlinks validation was seg faulting on the arm build because we only had the amd64 builds.  This appears to have been creating large .core dump files which were included in the final images.  I *believe* this should fix the issue, but am going to rethink the validation layer a bit next week.


abhay-krishna (approved):
/lgtm

PR #363, opened 2022-04-11
distro-base: fix ldd segfaults on arm builds

jaxesn:
*Issue #, if available:*

*Description of changes:*

The validate_libraries script uses ldd which for some reason seg faults under qemu in docker/buildkit.  These seg faults were causing core dumps which were getting included in the final images.

This moves the validation to its own build and only runs it on the image arch that matches the host machine.


eks-distro-base/Dockerfile.minimal-base-nginx
```diff
@@ -38,16 +38,11 @@ RUN set -x && \
     # remove unncessary utils and other binary deps that are not needed by nginx during runtime
     remove_package "chkconfig gawk grep info make openssl p11-kit p11-kit-trust sed shadow-utils" && \
     remove_package systemd true && \
+    # envsubst is needed to generate the config, but it comes from the gettext package
```
jaxesn, line 41: Small cleanup i did in other files in the last PR.

eks-distro-base/Dockerfile.minimal-helpers
```diff
@@ -26,6 +26,18 @@ FROM scratch as export
 COPY --from=builder /tmp/packages/* .
 
 
+################ Validate #######################
+ARG BUILT_BUILDER_IMAGE
+FROM --platform=$BUILDPLATFORM ${BUILT_BUILDER_IMAGE} as run_validate
```
jaxesn, line 31: This will only run this stage for the buildplatform instead of both.

abhay-krishna (approved):
/lgtm

PR #365, opened 2022-04-11
Add builder-base tag file to config updater list

abhay-krishna:
Add builder-base tag file to config updater list.


jaxesn (approved):
/lgtm

PR #378, opened 2022-04-26
builder-base: install go-licenses with 1.17 too

jaxesn:
*Issue #, if available:*

*Description of changes:*
Projects are now depending on 1.17 syntax which 1.16 can not parse.  I ran into this updating the kustomize-controller. The exact error is:

`github.com/hashicorp/vault/sdk/version: vendor/github.com/hashicorp/vault/sdk/helper/pluginutil/run_config.go:13:2: //go:build comment without // +build comment`

We have been always using go 1.16 when running go list and gather-licenses because of the way go-licenses finds the dependencies it needs to be installed via the version of go that is currently highest in the path when running.  Now that we have projects using 1.17 we need to support using go-licenses installed with both 1.16 and 1.17.


abhay-krishna (approved):
/lgtm

PR #379, opened 2022-04-26
base-git: bump libgit2 to 1.3.1

jaxesn:
*Issue #, if available:*

*Description of changes:*

source-controller is the only consumer of this base image and in their recent release they bumped libgit2 to 1.3.1. For reference, [this](https://github.com/fluxcd/golang-with-libgit2/blob/main/hack/static.sh) is the upstream script they use to create their build image.


abhay-krishna (approved):
/lgtm

PR #390, opened 2022-05-31
distro-base: Adds new install_binary scripts

jaxesn:
*Issue #, if available:*

*Description of changes:*

This PR sets up for future minimal images which are more tailored for the exact use case.  For ex, the ebs csi driver team has asked us to help them with a minimal image for them.  The ebs csi driver has a set of known exec's that it will exec out to and ideally we could build an image with those execs and their required libraries.  We could also look at using this approach for some of our existing images, like nginx + haproxy + docker-client, to clean up those dockerfiles. 

*Note* This PR only includes to the scripts to keep the PR as small as possible and properly document this new approach.  I have tested a ebs csi driver image with the ebs team using these scripts to validate it passes their e2e suite.

To review the existing design: https://github.com/aws/eks-distro-build-tooling/tree/main/eks-distro-base#design

The key new script here `install_binary` which takes one or more binaries as input.  For each binary supplied, `yum provides` is called to determine which yum package we should download to get the binary from.  This method seems fairly consistent, there have been a few gotchas related to needing the full path when calling provides or in some cases needing not a full path, both overall this seems like a reasonable option.

Once the rpm is determined, it is downloaded and extracted to a tmp folder.  The desired binary is then copied into place in the /newroot folder.  `ldd` is run against the binary to determine which libraries are needed in the resulting image.  For each of these libraries, yum provides is again used to determine the correct rpm to install.  These rpms are installed as is, instead of trying to limit them down, since in general lib rpms just have the expected libs.  A future enhancement could be to look at only pulling over the libraries that are needed.

For each rpm, be it fully installed or just certain binaries are pulled out, a record is created in the rpm db which is included in the final image.  By providing this rpm db in the image, scanners such as ECR can correctly identify the packages that make up the image.  

Once the image is created, there is separate "build" that runs which validates the installed binaries in the final image, using a similar ldd process to confirm the needed libraries are included.

EBS CSI: https://github.com/aws/eks-distro-build-tooling/pull/398
Docker client: https://github.com/aws/eks-distro-build-tooling/pull/399
Haproxy: https://github.com/aws/eks-distro-build-tooling/pull/400

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 
 .PHONY: run-buildkit-and-registry
 run-buildkit-and-registry:
-	docker run -d --name buildkitd --net host --privileged moby/buildkit:v0.9.3-rootless
+	docker run -d --name buildkitd --net host --privileged moby/buildkit:v0.10.3-rootless
 	docker run -d --name registry  --net host registry:2
 
 .PHONY: stop-buildkit-and-registry
--- a/eks-distro-base/Dockerfile.minimal-helpers
+++ b/eks-distro-base/Dockerfile.minimal-helpers
@@ -33,12 +33,17 @@ RUN set -x && \
     find . -type f -follow -printf '%s\t/%P\n' | numfmt --field=1 --to=iec-i --padding=8 --suffix=B | sort -k 2 > /tmp/packages/$VARIANT-files
 
 FROM scratch as export
-COPY --from=prepare-export /tmp/packages/* .
+
+ARG VARIANT
+
+COPY --from=prepare-export /tmp/packages/$VARIANT* .
 
 
 ################ Validate #######################
 ARG BUILT_BUILDER_IMAGE
 FROM --platform=$BUILDPLATFORM ${BUILT_BUILDER_IMAGE} as run_validate
+
+COPY scripts/ /usr/bin
 
 RUN set -x && \
     validate_libraries && \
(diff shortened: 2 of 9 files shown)


Makefile
```diff
@@ -1,7 +1,7 @@
 
 .PHONY: run-buildkit-and-registry
 run-buildkit-and-registry:
-	docker run -d --name buildkitd --net host --privileged moby/buildkit:v0.9.3-rootless
+	docker run -d --name buildkitd --net host --privileged moby/buildkit:v0.10.3-rootless
```
jaxesn, line 4: we use this in builder base now.

eks-distro-base/Dockerfile.minimal-helpers
```diff
@@ -40,6 +40,8 @@ COPY --from=prepare-export /tmp/packages/* .
 ARG BUILT_BUILDER_IMAGE
 FROM --platform=$BUILDPLATFORM ${BUILT_BUILDER_IMAGE} as run_validate
 
+COPY scripts/ /usr/bin
```
jaxesn, line 43: doing this here just helps iterating without having to rebuild the full image

eks-distro-base/scripts/clean_docs
```diff
@@ -16,15 +16,14 @@
 
 set -e
 set -o pipefail
-set -x
```
jaxesn, line 19: I think we can make these scripts a bit less noisy now, esp since we are tracking the full file list now...

eks-distro-base/scripts/cleanup
```diff
@@ -23,7 +23,7 @@ VARIANT="$1"
 clean_docs $NEWROOT
 
 clean_yum
-rm -rf $DOWNLOAD_DIR $NEWROOT/var/lib/yum
+rm -rf $DOWNLOAD_DIR $NEWROOT/var/lib/yum $NEWROOT/var/log/yum.log $NEWROOT/usr/share/bash-completion
```
jaxesn, line 26: small cleanup

abhay-krishna (approved):
/lgtm
/approve

PR #413, opened 2022-06-23
Introducing minimal-base-nsenter variant

alegermer:
### Summary
This adds a new minimal variant containing the `nsenter` utility from `util-linux` package. 

### Description
This base image is intended to be used by downstream implementations running on privileged containers when there is the requirement to gain access to the underlying host OS namespaces, which basically allows processes started from within the container to run as if they had been started from the parent host (outside the container's sandbox).

Because use cases that would explore this feature must run under privileged permissions it is of utmost importance that no other unnecessary dependencies are included on the image, to eliminate 3rd-party vulnerability risks.

The `nsetner` use is inspired by the [Bottlerocket sheltie script embedded on the bottlerocket-admin-container](https://github.com/bottlerocket-os/bottlerocket-admin-container/blob/develop/sheltie).

### Examples

```shell
# Getting dmesg from your local systemd-enabled host from within the container
docker run --rm -it --pid=host --privileged --name minimal \
  "public.ecr.aws/b8b1v8t5/eks-distro-minimal-base-nsenter:2022-06-22-1655945127" \
  /usr/bin/nsenter -t 1 --mount journalctl --dmesg --no-pager
```

```shell
# Listing /var/log path from your local host from within the container
docker run --rm -it --pid=host --privileged --name minimal \
  "public.ecr.aws/b8b1v8t5/eks-distro-minimal-base-nsenter:2022-06-22-1655945127" \
  /usr/bin/nsenter -t 1 --mount ls -lha /var/log/
```
----


abhay-krishna:
❤️

jaxesn (commented):
/ok-to-test

This looks pretty good to me. @abhay-krishna can you confirm the cdk change has rollout?

eks-distro-base/Dockerfile.minimal-base-nsenter
```diff
@@ -0,0 +1,32 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# *NOTE* we have to limit our number of layers heres because in presubmits there
+# is no overlay fs and we will run out of space quickly
+
+################# BASE-NONROOT ################
```
jaxesn, line 19: Super tiny nit and I suspect we have this issue in other files, but change this to base!-nsenter instead of base-nonroot

alegermer: Good catch! I have actually changed it to `## BUILDER ##`  to be consistent with most of the other Dockerfiles. The `csi-ebs` one is the one which has the same problem at this point. I didn't fix that one in this PR to not pollute its scope.

jaxesn:
2022 build is failing due to ebs csi, not sure why but we can ignore it and merge this anyway, I'll fix 2022 next week.

/override eks-distro-base-tooling-presubmit-2022

PR #417, opened 2022-07-05
Copy Tuftool to top-level bin directory

abhay-krishna:
So that it is accessible by all users.


pokearu (approved):
/approve
/lgtm

PR #426, opened 2022-07-19
distro-base: support updating al22 images if new releasever

jaxesn:
*Issue #, if available:*

*Description of changes:*

AL22 uses `releasever` which is different than al2.  Currently during their beta period, whenever they release a new build of al22 they are bumping the releasever and all new packages are under this version.  I think the best thing to do here is when there is a new releasever from al22 we should rebuild our images.  I want to watch what happens when they go ga and how often they are bumping new releasevers.

If they are going to be bumping the releasever more often than wed like to rebuild our minimal base images, the approach we can take is to get the latest releasever and pass that to `yum check --security --releasever=<>`. Well wait to see what happens post ga to do this.


jaxesn:
I think sometime in oct.

PR #436, opened 2022-07-28
distro-base: Various improvements to al22

jaxesn:
*Issue #, if available:*

*Description of changes:*

- AL22 images were not actually being built from the latest. The way al22 works with this releasever stuff is that yum upgrade doesnt actually get to the latest, if we do not set the new releasever.  This meant unless the AL22 team published new container images, which they do not always do, for each releasever would constantly rebuild our images because check update was correct but build was not.
- I found the above due to a build failure in the al22 postsubmit for the git variant.  This was due to the fact that a newer package from al22 had been released and what I tested with was on an image with an older releasever and therefore an older package.  A new bin had been added which we do not need so its been added to the unnecessary files list.
- Improve check_update a bit by installing the dnf plugin before copying the yum db from the minimal image
- Move the yum install perf improvement from the python branch, these changes are just indention changes no real code change other than moving the if !shallow up.
- Changed the update other repos pr script to skip updates where the new tag is `null`, which is what we typically do to trigger new builds.

### DIFF
--- a/eks-distro-base/Dockerfile.minimal-base
+++ b/eks-distro-base/Dockerfile.minimal-base
@@ -32,7 +32,13 @@ COPY scripts/ /usr/bin
 ENV DOWNLOAD_DIR /tmp/download
 
 ENV NEWROOT /
+
 RUN set -x && \
+    if grep -q "2022" "/etc/os-release"; then \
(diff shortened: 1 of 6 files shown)


abhay-krishna (approved):
/lgtm
/approve

PR #438, opened 2022-07-29
distro-base: add ability to be build dependent images

jaxesn:
*Issue #, if available:*

*Description of changes:*

When working locally i have wanted the ability to build an image along with the base image(s) it depends on.  What i usually do is `make minimal-images-base minimal-images-base-glibc minimal-images-base-iptables`, which is annoying. I also think we could use this for presubmits to break them up a bit for this job, esp as we add more images.

Getting the order right for the various targets has always been an issue with all the wildcard rules and ive never really understood why. When introducing the dep building support, it was easier to go ahead and define most targets as actual targets, using variables since they are generated, to make the targets more clearly defined.

/hold

### DIFF
--- a/eks-distro-base/check_update.sh
+++ b/eks-distro-base/check_update.sh
@@ -39,14 +39,18 @@ FROM public.ecr.aws/amazonlinux/amazonlinux:$AL_TAG as builder
 
 RUN set -x && \
     if grep -q "2022" "/etc/os-release"; then \
-        yum install -y dnf-plugin-release-notification && \
-        yum check-release-update &> /tmp/new-release && \
-        if grep "A newer release" /tmp/new-release; then echo "100" > ./return_value; fi \
+        yum install -y dnf-plugin-release-notification; \
     fi
 
 RUN rm -rf /var/lib/rpm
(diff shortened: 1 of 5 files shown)


eks-distro-base/check_update.sh
```diff
@@ -39,15 +39,19 @@ FROM public.ecr.aws/amazonlinux/amazonlinux:$AL_TAG as builder
...
-        if grep "A newer release" /tmp/new-release; then echo "100" > ./return_value; fi \
+        yum install -y dnf-plugin-release-notification; \
     fi
 
 RUN rm -rf /var/lib/rpm
 COPY --from=base_image /var/lib/rpm /var/lib/rpm
 COPY --from=base_image /etc/yum.repos.d /etc/yum.repos.d
 
+RUN set -x && \
+    if grep -q "2022" "/etc/os-release"; then \
```
jaxesn, line 50: this has to be run after replacing the yum db otherwise it wont necessarily see a new verison.

kschumy: Should we add that as a comment?

PR #443, opened 2022-08-02
gen-attribution: Handles multi case license file names

jaxesn:
*Issue #, if available:*

*Description of changes:*

We finally ran into the first case where a LICENSE file wasnt named upper case in https://github.com/aws/eks-anywhere-build-tooling/pull/1107.  The code previously looked for various versions of LICENSE(.md|txt) but still required them to be upper case.  This change introduces glob which allows for regex like dir navigating to support any casing of license.


abhay-krishna (approved):
/lgtm
/approve

PR #448, opened 2022-08-04
Adds nodejs runtime and compiler images

jaxesn:
*Issue #, if available:*

*Description of changes:*

This adds a new minimal runtime and compiler image for nodejs.  The intent is to use this image to containerize our generate attribution nodejs app used across eks-d and eks-a. I think this is also a reasonable addition to our set of compilers as we drive to adding even more to the minimal image set.

The main change here is the new dockerfile for nodejs.  The pattern followed is similar to the golang image where there is a layer which downloads nodejs which is cached between the different image builds so it only has to build once. This also has the sideffect (pro?) of subsequent builds of these images due to OS package updates will just use the previous build's cache which is stored inline in ecr.  The key difference to golang, and why there are more changes to the Makefile than I had hoped, is golang does not have a runtime image, its just the minimal-base, so the caching happens in the golang-compiler:base image, vs node which caches in the base-nodejs (runtime) image.

The first commit is the meat and the second are the test additions. No changes to really call out there, its just the new files for nodejs and a couple of the ones that "build" everything, all of which look good.


eks-distro-base/tests/run_tests.sh
```diff
@@ -318,4 +318,18 @@ check_base-python3.9() {
...
 }
 
+ check_base-nodejs18() {
+    for platform in ${PLATFORMS//,/ }; do
+        if docker run --platform=$platform --pull=always $IMAGE_REPO/eks-distro-minimal-base-nodejs:18-$IMAGE_TAG node --version | grep -v '18'; then
+            echo "nodejs issue!"
+            exit 1
+        fi
+
+        if docker run --platform=$platform --pull=always $IMAGE_REPO/eks-distro-minimal-base-nodejs:18-$IMAGE_TAG npm --version | grep -v '8'; then
```
kschumy, line 328: Why 8?

kschumy:
@jaxesn What's the status of this?

jaxesn:
> @jaxesn What's the status of this?

Quite active

eks-distro-base/Dockerfile.minimal-base-nodejs
```diff
@@ -0,0 +1,108 @@
...
+ARG COMPILER_VERSION=${VERSIONED_VARIANT}
+ARG IS_COMPILER=false
+
+################# DOWNLOAD ON GLIBC ######################
+# Download in the minimal base builder image to allow sharing result across compiler
+ # and runtime images. We also build the cache into the runtime image
+ # which includes the cache data for this layer and along with COPY --link
+ # this also us to ensure this layer is always reusable across the runtime + compiler
+ # builds without relying on the buildkitd local cache
+FROM ${MINIMAL_BASE_BUILDER_IMAGE} as nodejs-download
```
jaxesn, line 35: this is the layer that ends up getting cached in the runtime image and shared between it and the compiler images

eks-distro-base/Makefile
```diff
@@ -157,6 +158,23 @@ STANDARD_LATEST=$(foreach tag,$(if $(filter $(AL_TAG),2022),latest-al22,latest l
 COMPILER_VERSION_WITH_VARIANT=$(if $(BASE_VERSIONED_VARIANT),$(COMPILER_VERSION)-$(BASE_VERSIONED_VARIANT),$(COMPILER_VERSION))
 COMPILER_FULL_VERSION_WITH_VARIANT=$(if $(BASE_VERSIONED_VARIANT),$(COMPILER_FULL_VERSION)-$(BASE_VERSIONED_VARIANT),$(COMPILER_FULL_VERSION))
 COMPILER_LATEST=$(if $(filter $(AL_TAG),2022),$(COMPILER_VERSION_WITH_VARIANT)-al22 $(COMPILER_FULL_VERSION_WITH_VARIANT)-al22,$(COMPILER_VERSION_WITH_VARIANT) $(COMPILER_FULL_VERSION_WITH_VARIANT) $(COMPILER_VERSION_WITH_VARIANT)-al2 $(COMPILER_FULL_VERSION_WITH_VARIANT)-al2)
+
+# $1 - language short name (python/nodejs/golang/java)
```
jaxesn, line 162: these new macros is to handle determining if we pull the cache from the compiler-base image (in the case of golang) or the runtime image (like for nodejs and in the future python)

eks-distro-base/Makefile
```diff
@@ -285,58 +308,97 @@ define CREATE_PR_TARGET_BODY
...
 endef
 
-# $1 - language short name (python/nodejs/golang/java)
 define COMPILER_TARGET_OVERRIDES
-$(eval COMPILER_VARIANTS:=$(BASE_$(call TO_UPPER,$(1))_COMPILER_VARIANT_VERSIONS))
-$(eval VERSIONS:=$(BASE_$(call TO_UPPER,$(1))_VARIANT_VERSIONS))
-$(eval COMPILER_VERSIONED_VARIANTS:=$(ALL_MINIMAL_IMAGE_VERSIONED_VARIANTS_TARGETS_FOR_BASE_$(call TO_UPPER,$(1))_COMPILER))
-$(eval COMPILER_BUILDER_VERSIONED_VARIANTS:=$(BUILDER_MINIMAL_IMAGE_VERSIONED_VARIANTS_TARGETS_FOR_BASE_$(call TO_UPPER,$(1))_COMPILER))
-$(eval VARIANT_NAME:=$($(call TO_UPPER,$(1))_VARIANT_NAME))
+$(eval COMPILER_NAME:=$(1))
```
jaxesn, line 312: These are mostly variable renames.

markapruett (approved):

PR #456, opened 2022-08-23
Bump Golang version to 1.18.5 on builder-base

abhay-krishna:
Bumping Golang to v1.18.5 on builder-base.


kschumy:
We're force-merging this and hoping for the best. According to Abhay, "we'll need to fix the EOF issue before adding more tools to builder-base"

PR #474, opened 2022-09-22
golang 1.15 project directory and patches

danbudris:
*Issue #, if available:*
https://github.com/aws/eks-distro-internal/issues/206

*Description of changes:*
Initial commit for a Golang 1.15 build.

Includes the dir structure (projects/golang/go), makefile, and patches.

Currently, makefile is only set up to clone + patch + build, we're going to add the 'release' portion of the system later once we have this ironed out and have a distribution point decided on. It will probably initially be the postsubmits bucket.

Also, this is just using the native go tooling to build the binary, for now; will add the RPM build later.

This pairs with the presubmit in https://github.com/aws/eks-distro-prow-jobs/pull/341


danbudris:
will be pulling in aspects of https://github.com/aws/eks-distro-build-tooling/pull/472/files#diff-7a95faef450f7e6665c519e73e648c2b564dc4294ed9bdbb96651574477ab59c for RPM build and additional patches

projects/golang/go/Makefile
```diff
@@ -0,0 +1,38 @@
...
+
+.PHONY: clean
+clean:
+	rm -rf $(CLONED_REPO_DIRECTORY)
+
+.PHONY: check-env
+check-env:
+ifndef GO_SOURCE_VERSION
+$(error environment variable GO_SOURCE_VERSION, e.g. 1.15, is undefined)
+endif
```
rcrozean, line 38: Noticed this in many of our Makefiles: 
`########### DO NOT EDIT #############################`
`# To update call: make add-generated-help-block`
`# This is added to help document dynamic targets and support shell autocompletion`
`# Run make help for a formatted help block with all targets`
`include Help.mk`
`########### END GENERATED ###########################`

Is this an artifact from previous tools or is this something we still use?

projects/golang/go/1.15/patches/0005-go-1.15.15-eks-math-big-prevent-overflow-in-R.patch
```diff
@@ -9,7 +9,8 @@ Backported To: go-1.15.15-eks
 Backported On: Thu, 22 Sept 2022
 Backported By: budris@amazon.com
 Backported From: release-branch.go1.16
-Source Commit: https://github.com/golang/go/commit/07ee9e6445057e181fb3895df978748ffef30327
+Upstream Source Commit: https://github.com/golang/go/commit/07ee9e6445057e181fb3895df978748ffef30327
+EKS Patch Source Commit: https://github.com/danbudris/go/commit/f56e2b486b1cba8eae7326dac48da67c16ebba28
```
rcrozean, line 13: Is this for the patches that required differences between upstream an your commit? Should we be doing this format on all of our patches that differ? Mostly wondering if I should do it [here](https://github.com/aws/eks-distro/pull/1254) for the Kubernetes backports, or how this extends to other patching we are doing.

danbudris: I decided to include it on all of them in this specific project in order to ensure we're tracking everything that's done and can easily backtrack our steps if required as we get further into the testing phase. I think it'd be a reasonable to at least note in the k8s patch the changes that had to be made between the source commit and the patch, so that we have a record of what modifications we required

markapruett (approved):

projects/golang/go/Makefile
danbudris: this is something we still use, tho not everywhere

PR #477, opened 2022-09-27
Golang 1.15 RPM build

danbudris:
*Issue #, if available:*
https://github.com/aws/eks-distro-internal/issues/206

*Description of changes:*
make targets to build the golang RPMs from source.

the `build` target will now:
- set up the rpmbuild tree (`~/rpmbuild/{SOURCES,SPECS,BUILD,BUILDROOT,SRPMS}`
- copy the patches for the given version over to the rpmbuild tree
- copy the neccessary sources for the given version to the rpmbuild tree
- fetch the source for the given version from the upstream release tag
- execute the RPM build targeting the spec in `$VERSION/rpmbuild/SPECS`, with the resulting RPMs being output to a sub-dir of the version directory (e.g. `1.15/rpmbuild/RPMS/{x86_64,noarch}`)

The presubmit executes this build target, and each time we change the patches or modify the spec it'll run through this process to validate that the build works successfully.

Additionally, this PR adds in the 3 original patches from upstream as 101, 102, and 103, and modifys the spec to specifically apply all the patches from the source.


projects/golang/go/Makefile
```diff
@@ -1,38 +1,70 @@
...
+copy-patches-to-rpmbuild-tree:
+	cp -r $(VERSION_DIRECTORY)/patches/* $(VERSION_DIRECTORY)/rpmbuild/SOURCES/
+
+.PHONY: rpmbuild-golang
+rpmbuild-golang:
+ifeq ($(shell uname),Darwin)
+	$(error "rpmbuild is being run on a Mac, build this on an Amazon Linux machine!")
+endif
+ifeq (,$(shell which rpmbuild))
+	$(shell echo "No rpmbuild in $(PATH)")
```
jaxesn, line 48: id just error here, maybe output how to install it vs trying to install it.  the install wont work if sudo is required and running the whole make target as sudo doesnt seem like what wed want to do.

danbudris: ok, that seems fine. We can then just install rpmbuild on the builder base

projects/golang/go/Makefile
```diff
@@ -1,38 +1,70 @@
...
+.PHONY: fetch-golang-source-archive
+fetch-golang-source-archive:
+	curl -L -o $(VERSION_DIRECTORY)/rpmbuild/SOURCES/$(GIT_TAG).src.tar.gz https://github.com/golang/go/archive/refs/tags/$(GIT_TAG).tar.gz --create-dirs 
+
+.PHONY: copy-patches-to-rpmbuild-tree
+copy-patches-to-rpmbuild-tree:
+	cp -r $(VERSION_DIRECTORY)/patches/* $(VERSION_DIRECTORY)/rpmbuild/SOURCES/
+
+.PHONY: rpmbuild-golang
+rpmbuild-golang:
```
jaxesn, line 43: i think the most correct thing here would be to have the download and patches as pre-reqs here

projects/golang/go/Makefile
```diff
@@ -1,38 +1,74 @@
...
+endif
+ifeq(, $(shell which rpmbuild))
+	$(error "No rpmbuild in $(PATH), try 'yum install rpmbuild'")
+endif
+	rpmbuild -ba $(VERSION_DIRECTORY)/rpmbuild/SPECS/golang.spec --define "_topdir $(VERSION_DIRECTORY)/rpmbuild" --define "_tmpdir $(VERSION_DIRECTORY)/rpmbuild/tmp)" --define "_rpmdir $(VERSION_DIRECTORY)/rpmbuild"
+
 .PHONY: clean
 clean:
 	rm -rf $(CLONED_REPO_DIRECTORY)
+	rm -rf $(VERSION_DIRECTORY)/rpmbuild/noarch
```
jaxesn, line 55: should we just delete the whole dir, strong reason to call out each sub dir specifically?

projects/golang/go/Makefile
danbudris: added dep install as another target we can call in the prow jobs

projects/golang/go/Makefile
danbudris: the way it's set up now the `golang.spec` is just stored in `rpmbuild/SPECS`, and a couple things are stored in `rpmbuild/SOURCES`, like the gdbinit and some macro settings

danbudris: and we want to retain those

projects/golang/go/Makefile
danbudris: instead of having it all as pre-reqs of `build-golang-rpm`?

projects/golang/go/Makefile
jaxesn: how about making a working dir and copying all that to it so that then we can delete the whole thing when we are done?

danbudris: yeah, we could do that, we're doing that already with the patches (they're all in `patches/`) and copying them to `rpmbuild/SOURCES` at build-time

projects/golang/go/Makefile
danbudris: ok done

projects/golang/go/Makefile
danbudris: now that we're building from `$(HOME)/rpmbuild`, I'm a little hesitant to just clean that whole dir -- given one of us could have a set of rpm sources or specs other than golang hanging around. Thoughts @jaxesn

PR #490, opened 2022-10-04
golang 1.16 rpm spec, patches, and project structure

danbudris:
*Issue #, if available:*
https://github.com/aws/eks-distro-internal/issues/221
https://github.com/aws/eks-distro-internal/issues/207

*Description of changes:*
Add patches for go 1.16
Add RPM spec for go 1.16
Add folder structure for project overall

This has been tested by running the rpm build make target on an AL machine; I'm going to hold it for now untill we can run the pre-submits against this change (see https://github.com/aws/eks-distro-prow-jobs/pull/354) first and ensure it builds.


danbudris:
pre-submits passing; removed hold

projects/golang/go/Makefile
```diff
@@ -64,7 +64,7 @@ clean:
 
 .PHONY: install-deps
 install-deps:
-	yum update -y && yum install -y which git rpmdevtools awscli rpm-build golang-1.15.14-1.amzn2.0.1 pcre-devel glibc-static hostname procps-ng
+	yum update -y && yum install -y which git rpmdevtools go-srpm-macros awscli rpm-build golang-1.15.14-1.amzn2.0.1 pcre-devel glibc-static hostname procps-ng
```
jaxesn, line 67: i think for the 1.16 build we should install the 1.16 yum package? or from upstream if it doesn't exist?

projects/golang/go/1.16/patches/0101-syscall-expose-IfInfomsg.X__ifi_pad-on-s390x.patch
```diff
@@ -0,0 +1,41 @@
+From 2e472ee2cfd536a2f62c2c66be47dc831f524beb Mon Sep 17 00:00:00 2001
```
markapruett, line 1: Remind me why this one has a different structure than the other patches?

projects/golang/go/1.16/rpmbuild/SOURCES/golang-gdbinit
```diff
@@ -0,0 +1 @@
+add-auto-load-safe-path /usr/lib/golang/src/runtime/runtime-gdb.py
```
markapruett, line 1: What does this do?

danbudris: gdb is the GNU debugger; Golang includes tests for GDB compatibility. 

In GDB, you can load helper scripts which make it easier to convert your data structures to debugger calls and do other helpful things; in order to include these external helper scripts in the tests, you need to explicitly set their path and allow them to be imported safely.

This gdbinit file does that, and was included with the upstream RPM as a source; we're just carrying it over here

projects/golang/go/1.16/rpmbuild/SPECS/golang.spec
```diff
@@ -0,0 +1,1333 @@
...
+
+%global go_api 1.16
+%global go_version %{go_api}.15
+
+# For rpmdev-bumpspec and releng automation
+%global baserelease 1
+
+Name:           golang
+Version:        %{go_version}
+Release: %{baserelease}%{?dist}.0.1
```
markapruett, line 116: Do we want to have this hardcoded here?

projects/golang/go/1.16/patches/0101-syscall-expose-IfInfomsg.X__ifi_pad-on-s390x.patch
danbudris: yeah these are patches that were included by upstream in the RPM; they have a different format than our patches, which I prefaced with the version branch we used as the source. happy to change the format on these guys if you have suggestions

markapruett: Fine with me for now until we learn otherwise.

markapruett (approved):

projects/golang/go/Makefile
danbudris: will address this in a following PR

PR #534, opened 2022-10-21
use prod releases of EKS Go for builder-base

danbudris:
*Issue #, if available:*
https://github.com/aws/eks-distro-internal/issues/211
https://github.com/aws/eks-distro-internal/issues/241

*Description of changes:*
Update the builder-base to use the 'prod' released versions of EKS Go.

Also allow over-riding the source and version of the EKS Go versions installed. This will be used for testing Go versions.

Unify the output path for both Dev and Prod releaes of EKS Go

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #555, opened 2022-10-25
Break builder-base into multiple stages and layers

jaxesn:
*Issue #, if available:*

*Description of changes:*

This breaks install.sh into separate scripts, and layers in the dockerfile, per component included in the builder-base.  The end result is a mostly the same image, which a dozen layers vs 1 giant layer. Using this approach allows rebuilds locally to be a TON more manageable and it allows us to reuse the cache between different builds of builder-base, thus reducing build times even more.

Key changes:
- rust has been removed from the final builder base, install only tuftool is incldued
- go 1.14 has been removed, the final project we built 1.14 with was removed from eks-a not too long ago
- the instal golang script is currently not broken up by version, but as a follow up will be once @danbudris's work on custom golang builds is complete

I have diff'd the old image file list with a new image I built to confirm the only changes are the above expected ones.


builder-base/Makefile
```diff
@@ -59,7 +66,12 @@ images: buildkit-check
 		--frontend dockerfile.v0 \
 		--opt platform=$(BUILDKIT_PLATFORMS) \
 		--opt build-arg:BASE_IMAGE=$(BASE_IMAGE) \
+		--opt build-arg:BUILDER_IMAGE=$(BUILDER_IMAGE) \
 		--opt build-arg:GOPROXY=$(GOPROXY) \
+		--export-cache type=inline \
+		--import-cache type=registry,ref=public.ecr.aws/eks-distro-build-tooling/builder-base:$(CURRENT_BUILDER_BASE_TAG) \
```
jaxesn, line 72: This is new for us, but I think ok, basically in presubmit it will use the existing cache if it applies, much like the postsumit will... this has the down side of not building everytime, but the up side of not building everything every time.... considering post submit wont build everything every time either, assuming the cache applies, I think this is a reasonable tradeoff?

builder-base/Dockerfile
```diff
@@ -12,20 +12,141 @@
...
+ENV PATH="/go/bin/:$PATH"
+COPY ./scripts/install_golang.sh ./scripts/common_vars.sh \
+    ./checksums/golang-amd64-checksum ./checksums/golang-arm64-checksum /
+RUN export GOPROXY=${GOPROXY} && \
+    /install_golang.sh
+
+
+FROM ${BASE_IMAGE} as copy-stage
+
+# Ordered by least likely to be updated
```
jaxesn, line 118: This is fairly minor, probably only affects those of us who pull the builder base locally, but im open to opinions on the ordering here.  Since these are copy --link layers they do not depend on each other so cache can be reused between builds, however, the final layers would end up being different if any of the previous layers change, because the final layer has the parent layer hash in it.  If we make sure and put the layers that change most often towards the end then we can reuse the final layers more and more between builds.

danbudris: that sounds good to me. We know we'll be updating Go pretty frequently, so having that toward the front of the list makes sense. We can prolly revisit the order as needed if something should be moved up or down the list

builder-base/scripts/install_golang.sh
```diff
@@ -0,0 +1,171 @@
...
+}
+
+function build::go::install(){
+    # Set up specific go version by using go get, additional versions apart from default can be installed by calling
+    # the function again with the specific parameter.
+    local version=$1
+
+    # AL2 provides a longer supported version of golang, use AL2 package when possible
+    local yum_provided_versions="1.13"
+    local eks_built_versions="1.16.15 1.15.15"
```
danbudris, line 54: we've now got `1.17.13` and `1.18.7` builds as well

builder-base/scripts/install_golang.sh
```diff
@@ -0,0 +1,171 @@
...
+    if [[ $eks_built_versions =~ (^|[[:space:]])${version}($|[[:space:]]) && $TARGETARCH == "amd64" && $IS_AL22 == false ]]; then
+        local artifacts_bucket='eks-d-postsubmit-artifacts'
+        local arch='x86_64'
+        for artifact in golang golang-bin golang-race; do
+          curl https://$artifacts_bucket.s3.amazonaws.com/golang/go/go$version/RPMS/$arch/$artifact-$version-1.amzn2.0.1.$arch.rpm -o /tmp/$artifact-$version-1.amzn2.0.1.$arch.rpm
```
danbudris, line 59: we can update this to mirror: https://github.com/aws/eks-distro-build-tooling/pull/534/files, which has yet to be merged. This will point it to the 'prod'  artifacts for EKS Go

builder-base/scripts/install_golang.sh
```diff
@@ -0,0 +1,171 @@
...
+# go steps are not updated to support /newroot concept yet
+USR=/usr
+USR_LOCAL=/usr/local
+
+source $SCRIPT_ROOT/common_vars.sh
+
+VERSION="$1"
+
+# default, installed first
+GOLANG_VERSION="${GOLANG_VERSION:-1.18.6}"
```
danbudris, line 31: 1.18.7 I believe is latest

builder-base/scripts/install_golang.sh
danbudris: and `1.19.2` is open in: https://github.com/aws/eks-distro-prow-jobs/pull/382

abhay-krishna:
/lgtm
/approve
/hold

Awesome work @jaxesn, really excited for this! One more abstraction to consider would be to extract the common `wget -> shasum check -> extract and move to PATH` steps to a common function and use it where it applies, to reduce the duplication in scripts. Not sure if you want to do this in this PR or want to do it as a follow-up, so I'm approving and put it on hold to merge at your discretion.

jaxesn:
I had the same thought. I'm going to leave that for the future. One reason I didn't jump to that was to avoid introducing too many common scripts that could trigger rebuilds of every layer and instead optd for duplication to keep them as independent as possible. Worth considering more tho for sure!

PR #561, opened 2022-11-01
bump golang 1.18 to 1.18.8 and golang 1.19 to 1.19.3

danbudris:
*Issue #, if available:*

*Description of changes:*
Golang had patch version releases today. Update EKS Go to base off of the latest patch version releases.


projects/golang/go/1.18/rpmbuild/SPECS/golang.spec
```diff
@@ -108,7 +108,7 @@
 # Comment out go_prerelease and go_patch as needed
 %global go_api 1.18
 #global go_prerelease rc1
-%global go_patch 7
+%global go_patch 8
```
markapruett, line 111: What are these?

markapruett: NM. Got it.

markapruett (approved):

projects/golang/go/1.18/rpmbuild/SPECS/golang.spec
danbudris: patch version hard-coded in the spec file. we should change them to use a variable for it that we derive from the git-tag but for the moment I'm just bumping it

PR #566, opened 2022-11-03
use EKS Go relese version from the release trigger files if not speci…

danbudris:
…fied

*Issue #, if available:*

*Description of changes:*
Allow us to specify a specific EKS Go release version when installing EKS Go on the builder base.

If no release version is specified via the second argument to `build::go::install`, fetch the release number for the given Go version from the release trigger file in the repository.

### DIFF
--- a/builder-base/install.sh
+++ b/builder-base/install.sh
@@ -70,6 +70,9 @@ function build::go::install() {
   # the function again with the specific parameter.
   local version=$1
 
+  # EKS Go release version is the second positional argument
+  local eks_go_release=$2
+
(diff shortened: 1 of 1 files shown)


markapruett (approved):

danbudris:
talked to Jackson; we're going to take a slightly different approach here, and also wait till https://github.com/aws/eks-distro-build-tooling/pull/555 merges to work ont his.

jaxesn:
I think we can close this, we now have the tank file in place to define the go versions

PR #575, opened 2022-11-04
Rework makefile for distro-base and add dry run tests

jaxesn:
*Issue #, if available:*

*Description of changes:*

This reworks the makefile for eks-distro-base to allow for running make dry run tests.  This will allow us to ensure the buildctl commands that are being run are consistent between PR changes.  The main changes to the Makefile involve introducing "magic" vars to better support target overrides by full names vs using pattern matching and wildcards.  

There are no material changes to the resulting images, I have built them locally to confirm.

Commits:
- Refactor makefile to better support adding tests, also updated the pr scripts to allow overriding clone directories to make local testing easier, updates to check_updates to use buildkit script
- add help generation script, similar to eks-d and eks-a-build-tooling
- add dry run tests and expected files


eks-distro-base/make-tests/expected/base-csi-ebs/images/deps
```diff
@@ -0,0 +1,52 @@
+/override/root/../scripts/buildkit_check.sh
```
abhay-krishna, line 1: If these were generated files, can we add a comment to mark them as generated, so they are not edited inadvertently?

jaxesn: done!

PR #590, opened 2022-11-10
Add backport action

markapruett:
*Issue #, if available:*
https://github.com/aws/eks-distro/issues/1345

*Description of changes:*
Creates a github action that adds a /backport command. Tested on my fork.

When added to an issue this command will look at the MAINTAINED_EOL_VERSIONS file and open backports issue against the original. We could use Prow for this, but I wanted to see how difficult it was to do with GH Actions.


.github/workflows/backport.yml
```diff
@@ -0,0 +1,64 @@
...
+        title: ${{ env.ISSUE_TITLE }} - [eks go${{ matrix.go_version }} backport]
+        body: |
+          A backport to EKS Go ${{ matrix.go_version }} is [requested by ${{ env.REQUESTER }}](${{ env.HTML_URL }})
+        labels: security,golang
+env:
+  COMMENT: ${{ github.event.comment.body }}
+  ISSUE_URL: ${{ github.event.comment.issue_url }}
+  HTML_URL: ${{ github.event.comment.html_url }}
+  REQUESTER: ${{ github.event.comment.user.login }}
+  IS_BACKPORT: false
```
danbudris, line 63:
```suggestion
```

.github/workflows/backport.yml
```diff
@@ -0,0 +1,64 @@
...
+  IS_OWNER: false
```
danbudris, line 64:
```suggestion
```

.github/workflows/backport.yml
danbudris: I added these but they're not needed with this approach

PR #591, opened 2022-11-11
Adds github request caching to generate-attribution

jaxesn:
*Issue #, if available:*

*Description of changes:*

This is a small improvement specifically for the long running attribution job where all projects are generated. This adds a method to cache the responses from github and then save them out to a file in the users home directory.  If this file exists, it is loaded at startup and used instead of making the requests again.  This will probably cut off a decent amount of time of the overall job since this script isnt doing much other than making these requests and writing a file.


builder-base/checksums/nodejs-amd64-checksum
```diff
@@ -1 +1 @@
-faca6476cb5b41aa995370fd856d16fcfbbef5c18718a6fa44cc1bae4140849d  node-v16.18.0-linux-x64.tar.gz
+8949919fc52543efae3bfd057261927c616978614926682ad642915f98fe1981  node-v16.18.1-linux-x64.tar.gz
```
jaxesn, line 1: Bumping to force a builder base build.

PR #593, opened 2022-11-14
scaffolding for EKS Distro Build Tooling operations tool

danbudris:
*Issue #, if available:*
https://github.com/aws/eks-distro-internal/issues/250

*Description of changes:*
Scaffolding for an EKS Distro Build Tooling operational tool.

This adds:
- package structure
- makefile with example build targets for individual commands
- linter + unit test execution in makefile
- mocks support in Makefile
- logger cribbed from EKS-A

We can use this as the basis for further EKS-D Operational Tooling which requires us to go above-and-beyond shell scripting and make files.

Testing this out for yourself:
- check out this branch
- run `make build-eks-go-tools`
- run the eksGoTools binary that will be output to `./bin/darwin/$ARCH/eksGoTools` to see the command output

We can then flesh out the `pkg` with generalized operational tooling functionality (git & github interaction; ECR; S3; w/e) and build project and use-case specific cmds which utilize this generic functionality. 


rcrozean:
Is there a reason we shouldn't or couldn't move the pkg location up to something like tools/pkg? Just thinking in the long term other tools could be here and would use similar pkgs like the logger and github.

danbudris:
> Is there a reason we shouldn't or couldn't move the pkg location up to something like tools/pkg? Just thinking in the long term other tools could be here and would use similar pkgs like the logger and github.

so, I believe that `pkg` is in `tools/pkg`, actually

to add another tool to this that utilizes stuff in `pkg` (other than the go one I added in) we would just imitate the structure of the `eksGoTools` command, and call the fuctionality in pkg

danbudris:
related to: https://github.com/aws/eks-distro-build-tooling/pull/594

PR #595, opened 2022-11-14
Minimal cli scaffolding

rcrozean:
*Issue #, if available:*

*Description of changes:*
Just an initial structure for the minimal image cli. A few of the commands have been started getting defined. Will fit in to the path that is created with #593 


danbudris:
So, I'm wondering about the structure here -- it seems like it's leaning heavily on adding many individual commands.

For instance, do we need an `enabledDockerInstall` command attached to this? Or is it enough to add this as a flag to the root command which a user can specify? Same thing with the `enableExtra`, for example. My thought is these would be better suited as flags attached to specific commands or the root command, rather than their own individual commands. 

I was thinking we'd be able to do something like....

e.g. `build-minimal-image --config pathToMinimalImageConfig.yaml` 

...and wrap all of the other neccessary invocations under-the-hood of `build-minimal-image` or w/e we call it

Additionally, for the commands like `installRpm` and `installDepsForBinary` those strike me more as internal methods that will be invoked by a `build` command, for example, rather than top level commands themselves. I'd lean toward having a more limited command interface, and abstract as much of this away from the invoker as possible, moving it into PKG, for example.

tools/cmd/slim-jim/cmd/root.go
```diff
@@ -0,0 +1,44 @@
+package cmd
```
danbudris, line 1: let's make this a top-level remove command with sub commands in the same way as validate clean etc

danbudris (changes requested):

tools/minimalImageBuildTool/pkg/github/github_test.go
```diff
@@ -0,0 +1,44 @@
+package github_test
```
danbudris, line 1: Wondering, do we need the Github pr + issue creation functionality for the minimal images builder? It might be better to just not include the 'github', 'prManager', and 'issueManager' in the tool, since I don't think we'll actually need those for this specific use-case

PR #622, opened 2022-11-15
Uses versions.yaml to define builder base versions

jaxesn:
*Issue #, if available:*

*Description of changes:*
Had a conversation with @danbudris about automating the golang version updates. Switching to a yaml file should make that automation, as well as manual updates, much easier.


builder-base/Makefile
```diff
@@ -46,6 +46,10 @@ define BASE_TAG_FROM_TAG_FILE
 $(shell yq e ".al$(AL_TAG).$(1)" $(MAKE_ROOT)/../EKS_DISTRO_TAG_FILE.yaml)
 endef
 
+define NEWLINE
```
danbudris, line 49: what does this do?

jaxesn: check https://github.com/aws/eks-distro-build-tooling/pull/622/files#diff-af139d52c6b7cba3076bb8a1140743047f2567bc954a40a01f719fd16a679156R79

in make $shell removes newlines whatever is returned from the shell. For the most part this ends up being an ok decision, however in certain cases, like this one its kind of a pain.  Line 79 is using eval to convert the yaml file of keys and values to makefile variables, but when the newlines between each one are removed it creates a problem. You end up getting the first key being set to the all the other key and value pairs which is not what we want.  This is a common workaround for this issue to replace the newlines in the shell call to something else and then replace it back on the make side. Once that is done, the eval properly creates all the individual vars.

PR #627, opened 2022-11-17
Adds golang compiler images

jaxesn:
*Issue #, if available:*

*Description of changes:*

Add golang image builds for 1.15->1.19 using the compiler base/yum/gcc variants as base images.

I would focus on the first commit, the second commit is all the make test output files. I have spot checked them to ensure their accuracy. 


eks-distro-base/scripts/install_golang
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/env bash
```
jaxesn, line 1: This along with the versions in the makfile are duplicated from the builder-base build.  I am not sure i want to introduce a dep in the build process yet between these two so opting for copying for now until we see our update pattern and how much of a pain this is.  I also intend to remove the golang versions from the builder-base and switch to using these images directly, but that could be a longer term vision.

eks-distro-base/Makefile
```diff
@@ -240,6 +258,77 @@ define CREATE_PR_TARGET_BODY
 	)
 endef
 
+# $1 - language short name (python/nodejs/golang/java)
```
jaxesn, line 261: The following blob is the real meat of this PR, its very magicial. I tried to document where it made sense, please ask if anything isnt clear and I can improve the comments/code.

eks-distro-base/Makefile
```diff
@@ -136,6 +140,20 @@ PYTHON_VARIANT_NAME=eks-distro-minimal-base-python
...
+
+# ****************** GOLANG VARIANTS *************************
+GOLANG_VARIANT_NAME=eks-distro-minimal-base-golang
+BASE_GOLANG_VARIANT_VERSIONS=1.15 1.16 1.17 1.18 1.19
+
+GOLANG_1.15_FULL_VERSION=1.15.15-1
+GOLANG_1.16_FULL_VERSION=1.16.15-1
+GOLANG_1.17_FULL_VERSION=1.17.13-1
+GOLANG_1.18_FULL_VERSION=1.18.7-1
+GOLANG_1.19_FULL_VERSION=1.19.2-1
```
kschumy, line 152: Not sure if it matters, but the latest are 1.18.8 and 1.19.3

eks-distro-base/Makefile
kschumy: # 🪄

eks-distro-base/Makefile
jaxesn: I went with what we have the builder base currently. I figure we'll keep these two in sync

PR #628, opened 2022-11-18
Changes al22 check update to use --security and latest

jaxesn:
*Issue #, if available:*

*Description of changes:*

Al22 organizes packages into {releasever}s so with every package update you could pin to a specific set of packages. 
 This seems like a cool approach, but for us this is not exactly what we want since we are trying to minimize the number of updates we are actually doing to these images by only updating when there are security updates. They recently added a releasever latest to allow checking, and updating, across releasever without having to do a full system update.  This allows us to follow a similar approach as with al2 for checking for updates


abhay-krishna (approved):
/approve

PR #639, opened 2022-11-21
Switches to using go provided by eks-d minimal images

jaxesn:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/projects/golang/go/Makefile
+++ b/projects/golang/go/Makefile
@@ -81,7 +81,7 @@ clean:
 
 .PHONY: install-deps
 install-deps:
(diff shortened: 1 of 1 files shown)


projects/golang/go/Makefile
```diff
@@ -80,8 +80,15 @@ clean:
 	rm -rf $(HOME)/rpmbuild
 
 .PHONY: install-deps
+install-deps: T_GO_VERSION=1.15.15
+install-deps: T_GO_RELEASE_NUMBER=1
```
danbudris, line 84: any reason not to source this from the release trigger file for the given Go version?

danbudris: we'd need to account for the new prod release case but it'd be ideal if this wasn't hardcoded

PR #648, opened 2022-11-22
Adds python compiler to match golang style

jaxesn:
*Issue #, if available:*

*Description of changes:*

This PR adds a build for python 3.7, which required a couple of changes to better support building multiple compiler versions which have runtime images.

This also introduces the python compiler image and converts the existing image we add into a "runtime" image, basically just removes pip. I have built locally and compared the files export to confirm the expected difference. We can audit this when it builds in prow and checks in the new files.

*Note* When rolling this out to eks-a I will need to make a change to emissary build, but will handle this when that PR opens.

As usual, review by commit, the first being the most important. The other two are mostly just generated/test data.


danbudris (approved):

eks-distro-base/Dockerfile.minimal-base-python
```diff
@@ -20,15 +21,95 @@
...
+ARG IS_COMPILER=false
+
+
+################# COMPILE ON GLIBC ######################
+# Compile in the minimal base builder  image to allow sharing result across compiler
+# and runtime images. We also build the cache into the runtime image
+# which includes the cache data for this layer and along with COPY --link
+# this also us to ensure this layer is always reusable across the runtime + compiler
+# builds without relying on the buildkitd local cache
+FROM ${MINIMAL_BASE_BUILDER_IMAGE} as python-compile-3.9-al2
```
markapruett, line 37: Since it's set to `unused` above, how does this work? I see this in a couple places in this file.

jaxesn: Just ignored the word `unused`, thats a bad default. It is def used but due to a quirk with how docker parses these files, if that var isnt passed because its not needed for the given stage, it will still barf. Setting to something (anything) avoids it from bailing out.

markapruett (approved):

PR #650, opened 2022-11-23
Builds golang compiler by version instead of variant

jaxesn:
*Issue #, if available:*

*Description of changes:*

A small change to the ordering of golang builds. Instead of building all versions of go per compiler variant (base/yum/gcc) it now will build all the golang 1.15 versions then the 1.16 and so on.


danbudris (approved):
/lgtm

PR #651, opened 2022-11-23
Fixes base image tag when pulled from tag file

jaxesn:
*Issue #, if available:*

*Description of changes:*

Yesterdays periodic failed because the compiler-base images were not built, since they didnt need to be, but when attempting to build the golang images it ended up trying to use an improper tag for the compiler base images.  This was due to the version variant being prepended to the base_image_tag even when it was pulled from the file, which we do not want.  

There was a test that would have caught this, but I had missed it when adding the original expected output files.


eks-distro-base/make-tests/expected/base-golang-compiler-1.18-gcc/base-image-tag/use-tag-from-file
```diff
@@ -109,18 +133,24 @@ cat pushed file /override/root/eks-distro-minimal-base-compiler-gcc-pushed
...
+/override/root/../scripts/buildkit.sh build --frontend dockerfile.v0 --opt filename=Dockerfile.minimal-base-golang --opt platform=linux/amd64 --opt build-arg:OUTPUT_DEBUG_LOG= --opt build-arg:BASE_IMAGE=localhost:5000/compiler-base:some-previous-tag  --opt build-arg:BUILT_BUILDER_IMAGE=localhost:5000/golang-builder:1.18-gcc-test-tag --opt build-arg:VERSIONED_VARIANT=1.18-gcc --opt build-arg:AL_TAG=2 --opt build-arg:COMPILER_VERSION=1.18 --opt build-arg:COMPILER_FULL_VERSION=1.18.7-1 --opt build-arg:IS_COMPILER=true --progress plain --local dockerfile=./ --local context=. --opt target=final --export-cache type=inline --import-cache type=registry,ref=localhost:5000/golang-builder:some-previous-tag --opt no-cache=builder-golang-base --output type=image,oci-mediatypes=true,\"name=localhost:5000/golang:1.18-gcc-test-tag,localhost:5000/golang:1.18-gcc,localhost:5000/golang:1.18.7-1-gcc,localhost:5000/golang:1.18-gcc-al2,localhost:5000/golang:1.18.7-1-gcc-al2\",push=true
```
jaxesn, line 153: This diff is the proof!

PR #667, opened 2022-12-06
`golang-debian` base image for building upstream toolchain

danbudris:
*Issue #, if available:*

*Description of changes:*
The upstream Kubernetes build toolchain is composed of a set of cascading images built off of a Debian bullseye Golang builder image. While EKS Distro builds and maintains the entire supply-chain for our own AL2-based toolchain, we also need to provide some consumers a Debian-based Golang builder image for use with upstream toolchain tools like `kube-cross`. 

This PR adds a build for a Debian-based builder-base image which is built as part of the EKS Go build and release process.

The Dockerfile is based on [an upstream Golang dockerfile](https://github.com/docker-library/golang/blob/30403f1c144bf7773508cfbab5de09ecf4dbddf9/1.19/bullseye/Dockerfile), with mods and optimizations put in place based on our build and CI system. 

This build will now:
- build a `golang-debian` image, copying into the image the proper version of `EKS Go`
- if in a prod release, push that image to the public ECR for external consumption


projects/golang/go/Makefile
```diff
@@ -20,19 +20,32 @@ ifeq ($(ARCHITECTURE), ARM64)
 endif
 ARCH_LOWER=$(shell echo $(ARCHITECTURE) | awk '{print tolower($0)}')
 
-
 ARCH_RPM_OUT_PATH?=golang-$(GIT_TAG)/releases/$(BUILD_ID)/RPMS/$(RPM_ARCH)
 NOARCH_RPM_OUT_PATH?=golang-$(GIT_TAG)/releases/$(BUILD_ID)/RPMS/noarch
 ARCHIVES_OUT_PATH?=golang-$(GIT_TAG)/releases/$(BUILD_ID)/archives
 
+AWS_ACCOUNT_ID?=$(shell aws sts get-caller-identity --query Account --output text)
```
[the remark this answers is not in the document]
danbudris: yep, exactly -- so that the account ID will be set to the user who is making it

so when we construct 'IMAGE_REPO' [on 30 ](https://github.com/aws/eks-distro-build-tooling/pull/667/files/5247bec170ca07546ee98ff145561a8dd8fe9e2a#diff-5f1e44500fb103dc0ce96c63ccc7dca3b6dbceb62e74976286539988dd20f4a6R30) it'll be the ECR in the callers account

projects/golang/go/Makefile
```diff
@@ -121,6 +154,14 @@ ifndef GO_SOURCE_VERSION
 	$(error environment variable GO_SOURCE_VERSION, e.g. 1.15, is undefined)
 endif
 
+.PHONY: set-env
```
jaxesn, line 158: I dont believe this is going to work like you want. Each shell execution in make is a new shell. So even tho this one exports the path, it wont still be in affect by the time rpmbuild is called

jaxesn: I would probably jsut simplify this because we know we are builder from the builder base so we could just add this at the top of the makefile

`export PATH:=/go/goGO_SOURCE_VERSION/bin:$(PATH)`

danbudris: ✅

projects/golang/go/Makefile
```diff
@@ -107,7 +143,7 @@ clean:
 
 .PHONY: install-deps
 install-deps:
-	yum install -y which git rpmdevtools go-srpm-macros awscli rpm-build pcre-devel glibc-static hostname procps-ng tar
```
danbudris, line 110: now that we're using the builder-base as the base image for the EKS Go prow jobs we no longer need to install so many dependencies during the build. The `yum-plugin-ovl` is required for installing what dependencies we do have on the overlay fs.

projects/golang/go/docker/debianBase/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+ARG GOLANG_ARCHIVE_PATH
+
+ADD $GOLANG_ARCHIVE_PATH /usr/local
+
+ENV GOPATH /go
+ENV PATH /usr/local/go/bin:$PATH
+ENV PATH $GOPATH/bin:$PATH
+ENV GOLANG_VERSION GOLANG_VERSION
+
+RUN set -eux; \
```
danbudris, line 12: we're running this all as a single `RUN` command to avoid space-issues in pre-submit; if we have too many layers we'll run into disk space issues on the fargate builder nodes

jaxesn (approved):

PR #696, opened 2022-12-13
split prod release and prod image build into two seperate targets

danbudris:
*Issue #, if available:*

*Description of changes:*
Due to the limitations of service account for pods and the way that service account tokens are mounted, we are going to split the `images` and `release` into two make targets and two prow jobs; one prow job for building + pushing the images, one prow job for building + pushing the other artifacts. This way we can continue to ensure that each set of creds maintains least-priv.

These changes will be executed in two different prow jobs, as seen here https://github.com/aws/eks-distro-prow-jobs/pull/437


jaxesn (approved):
Seems reasonable

PR #699, opened 2022-12-13
Adds distro-base build for windows images

jaxesn:
*Issue #, if available:*

*Description of changes:*

We recently added windows builds for a couple of the csi sidecars in [eks-d](https://github.com/aws/eks-distro/pull/1494).  These builds are based off upstream's builds which use the nanoserver as the base image but pull the netapi dll from servercore.  The nanoserver image is actually a reasonable size, but the servercore image is a few gb compressed.  When trying to use that image in a presubmit (fargate) pod the extracted size ends up being over 9 gb!

Add a new windows base here would allow us to avoid downloading the extra large servercore in all of eks-d builds, instead we can limit to just the nanoserver and *maybe* even run these eks-d builds on the presubmit cluster vs now we run on postsubmit.

I have reused some of the manifest handling code from eks-d since we need that here to add the os.version to our built base.

The check-update is also a new process since we can not (as easily) actually run the container.  Instead this adds a flow to check the buildinfo in the pushed manifest and comparing the nanoserver/servercore sha's from that buildinfo against what is currently pushed upstream.  If they are different, a new image will be built.


eks-distro-base/check_update_windows.sh
```diff
@@ -0,0 +1,70 @@
...
+DIGEST_SHA=$(retry docker buildx imagetools inspect $BASE_IMAGE@$IMAGE_SHA --raw | jq -r '.config.digest')
+BUILDINFO=$(retry docker buildx imagetools inspect $BASE_IMAGE@$DIGEST_SHA --raw | jq -r '."moby.buildkit.buildinfo.v1"' | base64 -d)
+
+for variant in "nanoserver" "servercore"; do
+    SOURCE_IMAGE_REF=$(jq -r ".sources | .[] | select(.ref | contains(\"$variant\")) | .ref" <<< "$BUILDINFO")
+    SOURCE_IMAGE_PIN=$(jq -r ".sources | .[] | select(.ref | contains(\"$variant\")) | .pin" <<< "$BUILDINFO")
+    
+    if [ "$(retry docker buildx imagetools inspect --raw $SOURCE_IMAGE_REF@$SOURCE_IMAGE_PIN)" != "$(retry docker buildx imagetools inspect --raw $SOURCE_IMAGE_REF)" ]; then
+        echo "updates" > ./check-update/${NAME_FOR_TAG_FILE}  
+    fi
```
markapruett, line 67: Do we need an `exit 0` here to ensure the `echo` on the last line doesn't overwrite the 'updates' on line 66?

jaxesn: yes! adding

eks-distro-base/update_base_image.sh
```diff
@@ -24,8 +24,12 @@ IMAGE_NAME=$2
 AL_TAG=$3
```
markapruett, line 24: NAB: rename this to `TAG` instead?

markapruett (approved):

eks-distro-base/update_base_image.sh
jaxesn: let me think about that, its a good call but the AL_TAG concept is everywhere so id like to change it completely if going to do that.

PR #720, opened 2022-12-15
WIP - Add single version file for golang supported versions

markapruett:
*Issue #, if available:*

*Description of changes:*
Example of file we could use as the source of truth on the versions we support. It basically replicates what we have between the version dir and the GIT_TAG file, but what it adds is whether it's an in-support or out of support upstream version, which would be nice to have. 


markapruett:
I guess this isn't ideal because Prow can't trigger on the content.

PR #728, opened 2022-12-19
Fix EKS Go 1.16.15 CVE-2022-41716 patch; add standard lib install as smoke test

danbudris:
*Issue #, if available:*

*Description of changes:*
Our patch for CVE-2022-41716 against Go v1.16.15 was broken by the erroneous inclusion of an intermediate commit. This was not detected as windows functionality was not exercised during the compilation process and standard unit tests, and no checks were in place for windows functionality.

I've re-built and tested the patch; additionally, I've added `std` install for multiple GOOS/GOARCH combos in order to act as a smoke-test against this kind of issue in the future.


danbudris:
below pre-submits are encountring space issues on the prow cluster

/override golang-1-19-tooling-presubmit 
/override golang-1-18-tooling-presubmit

PR #735, opened 2022-12-20
Adds image mirror for pushing base images to public ecr

jaxesn:
*Issue #, if available:*

*Description of changes:*

This changes our base image builds to push to to private ecr repos as a staging area before updating the public ecr repos.  This will give us a chance to explictly decide when to push new images to public ecr, for public consumption.  For example, if a build were kick off that was not intended or produced a bad image for whatever reason, we have a chance to catch this vs now it would go straight to public ecr taking over the "latest" tag, which people do rely on.

The latest tags have also been removed (youll notice the test changes in commit 3) during the build process and instead only applied during the skopeo copy phase to the final public ecr repo.

The final postsubmit which triggers from the EKS_DISTRO_TAG_FILE.yaml changes (eks-distro-base-tag-files-update-postsubmit) will now run a `skopeo copy` for every image/tag combo.  The new code added to the makefile mirrors the same flow as the actual build, but instead of running buildctl it runs skopeo, the needed make vars are set for both targets so things like image_component/latest/etc are properly aligned.

/hold

Prow job changes: https://github.com/aws/eks-distro-prow-jobs/pull/448


markapruett:
Awaiting ecr creation before this PR is merged.

PR #744, opened 2022-12-21
Fix tag update logic for EKS-A Prowjobs

abhay-krishna:
Without the dash, the entirety of the tag was being selected for regex grouping, which caused an incorrect substitution.


kschumy (approved):
/lgtm

markapruett (approved):

PR #745, opened 2022-12-22
bump go versions in distro base makefile

danbudris:
*Issue #, if available:*

*Description of changes:*


jaxesn (commented):
We need to set the golang image tags in the tag file to null so the next periodic will do an actual build.

jaxesn (approved):

jaxesn:
The make tests fail due to the expected files having the go versions in them. I'm going to override this failure and fix those in another or to use a mockd version to avoid having to update these files when bumping versions

/override eks-distro-base-test-presubmit

jaxesn:
Oh yeah the windows one is also booked, overriding it as well. I have a fix on my other pr

/override eks-distro-base-tooling-presubmits-windows

PR #750, opened 2022-12-23
Cleans older buildkit deployments left from failed jobs

jaxesn:
*Issue #, if available:*

*Description of changes:*

The trap cleanup seems to work pretty well, there are only 2 left over right now that are from crashed jobs, but just to have a fail safe for these cases.  This approach is similar to what we do in eks-distro [postsubmits](https://github.com/aws/eks-distro/blob/main/development/kops/run_all.sh#L40) to clean up potentially old clusters that werent properly cleaned

### DIFF
--- a/scripts/setup_buildx.sh
+++ b/scripts/setup_buildx.sh
@@ -44,15 +44,7 @@ function setup_kubeconfig_in_pod() {
     kubectl config use-context cfc
 }
 
-if ! docker buildx version > /dev/null 2>&1; then
-    # TODO move to builder base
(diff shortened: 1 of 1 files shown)


scripts/setup_buildx.sh
```diff
@@ -44,15 +44,7 @@ function setup_kubeconfig_in_pod() {
     kubectl config use-context cfc
 }
 
-if ! docker buildx version > /dev/null 2>&1; then
```
jaxesn, line 47: this is in the builder base now

danbudris (approved):

PR #757, opened 2022-12-30
Trigger label with slash command

markapruett:
*Issue #, if available:*

*Description of changes:*
[Adding the 'do-not-merge/hold' label with the gh api ](https://github.com/aws/eks-distro-build-tooling/blob/main/pr-scripts/create_pr.sh#L128)doesn't work as the PR bot is not listed as an outside collaborator to our org (and just treated like an external developer). We could get around adding the bot by using a GH Action Labeler to hold all PRs from the bot, but even simpler is to use our existing Prow chatops feature with the slash hold command. 

Tested manually by adding the slash as the bot and it succceeded in adding hte label.

### DIFF
--- a/pr-scripts/create_pr.sh
+++ b/pr-scripts/create_pr.sh
@@ -79,6 +79,10 @@ else
     fi
 fi
 
+# Adding this here to include the "do-not-merge/hold" label. Trying to use the gh client with the --label arg will not succeed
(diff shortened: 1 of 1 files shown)


danbudris (approved):

abhay-krishna (approved):
/approve

PR #760, opened 2023-01-06
WIP: EKS Go and Builder Base codebuild support

danbudris:
*Issue #, if available:*
https://github.com/aws/eks-distro-internal/issues/308

*Description of changes:*
Codebuild build support for EKS Go and `builder-base` + dev EKS Go builds. Tested working in preprod pipelines.

The `buildspec*.yaml` files dictate what each stage in the pipeline executes.

The changes to the makefiles allow us to run in a codebuild environment, skip certain steps, and control other settings (like what cache to use, wether to build the Go images in preprod steps, etc).

The ECR credential helper and corresponding config are used to auth to ECR inside of the codebuild environment.


projects/golang/go/build/conformanceTests/eksDistroRebuildBuildspec.yaml
```diff
@@ -0,0 +1,11 @@
+version: 0.2
```
danbudris, line 1: this file is just a stub; it'll evolve as we work on the eks-distro rebuild system.

projects/golang/go/build/conformanceTests/builderBaseRebuildBuildspec.yaml
```diff
@@ -0,0 +1,22 @@
...
+    BUILDKIT_PLATFORMS: "linux/amd64"
+    GO_SOURCE_VERSION: "1.16"
+    IMAGE_NAME: "builder-base-development"
+    IMPORT_CACHE_REPOS: ""
+    UPDATE_BASE_IMAGE: "false"
+
+phases:
+  build:
+    commands:
+    - export GOLANG_RPM_DOWNLOAD_TARGETS=$(yq  e "to_entries | .[] | select(.key == \"*GOLANG_VERSION*\") | select(.value != \"$GO_SOURCE_VERSION*\") | .value |= \"download-golang-\" + . | .value" $CODEBUILD_SRC_DIR/builder-base/versions.yaml)
```
danbudris, line 15: this over-rides the golang versions we download to NOT download the one we're building in the previous pipeline step; that way, we skip the download, and just copy the RPMs over from the previous step (lines 19 and 20) prior to building the images

projects/golang/go/build/conformanceTests/builderBaseRebuildBuildspec.yaml
```diff
@@ -0,0 +1,22 @@
+version: 0.2
+
+env:
+  variables:
+    ARTIFACTS_BUCKET: "eks-d-postsubmit-artifacts"
+    BUILDKIT_PLATFORMS: "linux/amd64"
+    GO_SOURCE_VERSION: "1.16"
```
danbudris, line 7: this is going to change -- right now, I just have one pipeline which builds based on these buildspecs. However, before this is all wrapped up, we'll do this differently -- either by having a pipeline PER GO VERSION or having a pipeline with a build project per Go version. This env var will then be set in those pipeliens/proejcts.

config/docker-ecr-config.json
```diff
@@ -0,0 +1,8 @@
+{
```
danbudris, line 1: this is used by the ECR credentials helper to make authenticating to ECR inside of a codebuild job easier.

builder-base/Makefile
```diff
@@ -5,6 +5,12 @@ SHELL=bash
 .SUFFIXES:
 
 BASE_DIRECTORY=$(shell git rev-parse --show-toplevel)
+
+# In a codebuild environment the source is not a git repo, but simply a directory.
+ifdef CODEBUILD_SRC_DIR
```
jaxesn, line 10: why isnt a git repo?  Im just trying to understand the diff of the setup between this and eks-a-build-tooling?

projects/golang/go/Makefile
```diff
@@ -55,9 +65,10 @@ endif
 golang-build: check-env setup-rpm-tree fetch-golang-source-archive copy-sources-to-rpmbuild-tree copy-patches-to-rpmbuild-tree build-golang-rpm generate-golang-archive test
 
 .PHONY: build
-build: golang-build sync-artifacts-to-s3-dry-run local-images
-build:
-	echo "test"
+build: golang-build sync-artifacts-to-s3-dry-run
+ifeq ($(BUILD_PREPROD_IMAGES),true)
```
jaxesn, line 69: i would prefer not using ifeq, instead use a inline if in the build target above or setup a var at the top defining these targets, similar to https://github.com/aws/eks-anywhere-build-tooling/blob/main/Common.mk#L401

projects/golang/go/build/conformanceTests/builderBaseRebuildBuildspec.yaml
```diff
@@ -0,0 +1,22 @@
...
+  build:
+    commands:
+    - export GOLANG_RPM_DOWNLOAD_TARGETS=$(yq  e "to_entries | .[] | select(.key == \"*GOLANG_VERSION*\") | select(.value != \"$GO_SOURCE_VERSION*\") | .value |= \"download-golang-\" + . | .value" $CODEBUILD_SRC_DIR/builder-base/versions.yaml)
+    - scripts/setup_ecr_credential_helper.sh
+    - mkdir -p builder-base/tmp/golang-downloads
+    - make -C builder-base $GOLANG_RPM_DOWNLOAD_TARGETS
+    - cp ../01/projects/golang/go/$GO_SOURCE_VERSION/rpmbuild/x86_64/golang*.rpm $CODEBUILD_SRC_DIR/builder-base/tmp/golang-downloads
+    - cp ../01/projects/golang/go/$GO_SOURCE_VERSION/rpmbuild/noarch/golang*.rpm $CODEBUILD_SRC_DIR/builder-base/tmp/golang-downloads
+    - make -C builder-base copy-generate-attribution
+    - make -C builder-base images
```
jaxesn, line 22: doesnt images already run the copy-generate-attribution?

projects/golang/go/Makefile
danbudris: yeah, happy to do it that way. other than consistency, any reason you don't like the make `ifeq`?

projects/golang/go/build/conformanceTests/builderBaseRebuildBuildspec.yaml
danbudris: no -- `build` and `release` do. I'm running `copy-generate-attribution` and `images` independently in order to skip the checksum generation and validation

builder-base/Makefile
danbudris: took a look at this -- looks like I can tweak a setting, "full clone mode", that could allow the source repo to be a full git clone rather than the default, which is just a directory structure unpacked from a tarball. I'll look at updating the sources to support that

projects/golang/go/Makefile
danbudris: done

builder-base/Makefile
danbudris: ok yeah it's a bit more than a tweak -- the github source for codebuild doesn't support 'full clone mode' out of the box and it's going to require some modifications. Given how much we use the git metadata to orient ourselves in the repositoriy etc I'm going to take the time to set this up properly to pull from git rather than just copy over the source files in a tarball without the git metadata

danbudris: fixed -- as discussed, we'll just run a `git init` in the buildspec for now to initialize the directory structure.

PR #685, opened 2022-12-12
add golang checksum

zafs23:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #776, opened 2023-01-17
Download and Verify Golang checksum

zafs23:
*Issue #, if available:*

*Description of changes:* Golang checksums has been added for golang 1.15 - 1.19  for rpms and archives in [this PR](https://github.com/aws/eks-distro-build-tooling/pull/685). Now, in the current PR these files are downloaded and verified against the checksums. The acceptance criteria is tracked in [this PR](https://github.com/aws/eks-distro-internal/issues/314). 


projects/golang/go/Makefile
```diff
@@ -69,8 +69,23 @@ prod-release: golang-build sync-artifacts-to-s3
 prod-release-images: fetch-eks-go-archive images
 
 .PHONY: fetch-eks-go-archive
-fetch-eks-go-archive:
-	curl -L -o $(VERSION_DIRECTORY)/archives/$(GOOS)/$(ARCH_LOWER)/$(GIT_TAG).$(GOOS)-$(ARCH_LOWER).tar.gz https://distro.eks.amazonaws.com/golang-$(GIT_TAG)/releases/$(BUILD_ID)/archives/$(GOOS)/$(ARCH_LOWER)/$(GIT_TAG).$(GOOS)-$(ARCH_LOWER).tar.gz --create-dirs
+fetch-eks-go-archive: download-go-archive validate-go-archive-checksum
+
+.PHONY: download-go-archive
+download-go-archive:
+	curl -L -o $(VERSION_DIRECTORY)/archives/linux/amd64/go1.19.5.linux-amd64.tar.gz https://distro.eks.amazonaws.com/golang-go1.19.5/releases/1/archives/linux/amd64/go1.19.5.linux-amd64.tar.gz --create-dirs
```
jaxesn, line 76: Why did you remove the var use here, git_tag/build_id/arch/etc?

zafs23: @jaxesn not a final version, was testing locally without the vars.

zafs23: @jaxesn only 1.19 had the checksums added in S3 at that time.

projects/golang/go/Makefile
```diff
@@ -69,8 +69,23 @@ prod-release: golang-build sync-artifacts-to-s3
...
+	curl -L -o $(VERSION_DIRECTORY)/archives/linux/amd64/go1.19.5.linux-amd64.tar.gz.sha256 https://distro.eks.amazonaws.com/golang-go1.19.5/releases/1/archives/linux/amd64/go1.19.5.linux-amd64.tar.gz.sha256 --create-dirs
+
+.PHONY: validate-go-archive-checksum
+validate-go-archive-checksum:
```
danbudris, line 80: a trick we can use here to more easily support both archive and rpm checksums using the same make target is the `Make` operator `%`, to create a ["Pattern Rule"](https://www.gnu.org/software/make/manual/html_node/Pattern-Intro.html).

The `%` in target names acts like a wildcard, and you can then use the value of that wildcard inside the target.

So, for example, if we wanted to have a single `make` target `validate-go-%s-checksum`, we could call it for both `validate-go-archive-checksum` as well as `validate-go-x86-64-rpm-checksum`; and based on the value of the wildcard (the `%`) we decied what path to look in for the given files.

This way, we don't have to repeat lines 83 through 88 for each type of file we want to validate the checksum for.

For an example of what this might look like, check out [the `test-platform` target in this makefile](https://github.com/aws/eks-distro-build-tooling/blob/main/projects/golang/go/Makefile#L195). This allows us to have a single target, `test-platform-%`, which matches against any value like `test-platform-linux` or `test-platform-darwin`, without having to repeating the test code for each and every one of them

danbudris:
https://github.com/orgs/aws/projects/23/views/26?pane=issue&itemId=17870617

builder-base/scripts/download_golang.sh
```diff
@@ -33,19 +33,41 @@ function build::go::download(){
...
     done
 
     for artifact in golang-docs golang-misc golang-tests golang-src; do
         local filename="$outputDir/$artifact-$version-$RELEASE_NUMBER.amzn2.eks.noarch.rpm"
         if [ ! -f $filename ]; then
             curl -sSL --retry 5 https://distro.eks.amazonaws.com/golang-go$version/releases/$RELEASE_NUMBER/$arch/RPMS/noarch/$artifact-$version-$RELEASE_NUMBER.amzn2.eks.noarch.rpm -o $filename
         fi
+
+        local shaCheckFilename="$outputDir/tmp/$artifact-$version-$RELEASE_NUMBER.amzn2.eks.$arch.noarch.rpm"
+        curl -sSL --retry 5 https://distro.eks.amazonaws.com/golang-go$version/releases/$RELEASE_NUMBER/$arch/RPMS/noarch/$artifact-$version-$RELEASE_NUMBER.amzn2.eks.noarch.rpm -o $shaCheckFilename
```
danbudris, line 61: so, we don't want to perform the checksum check if the above check takes place-- the `! -f` indicates that if that file exists, we skip the download; we do this so that we can over-ride the golang RPM used with a local one. In this case, we don't want to perform the checksum check, as we're never using the downloaded RPM, but just using a local one.

builder-base/scripts/download_golang.sh
```diff
@@ -33,19 +33,41 @@ function build::go::download(){
...
+        curl -sSL --retry 5 https://distro.eks.amazonaws.com/golang-go$version/releases/$RELEASE_NUMBER/$arch/RPMS/noarch/$artifact-$version-$RELEASE_NUMBER.amzn2.eks.noarch.rpm.sha256 -o $shaCheckFilename.sha256
```
danbudris, line 62: why don't we just move the file we download if it's present, rather than re-download it? if we re-download it we're also not getting any guarantees about the integrity of the downloaded file above, just the one downloaded to tmp....and the call above in the `if` statement could mis-match but we wouldn't necessarily catch it

danbudris (approved):

danbudris (approved):

PR #778, opened 2023-01-17
Add golang linter and unit-test for EKS-D ops tool as GH actions

zafs23:
*Issue #, if available:*

*Description of changes:*


.github/workflows/go-coverage.yml
```diff
@@ -0,0 +1,29 @@
+name: Test coverage
+on:
+  push:
+    tags:
+      - v*
```
danbudris, line 5: I think we can remove the `tags` section, I'm not sure we care if this runs on tags

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,33 @@
+name: golangci-lint
+on:
+  push:
+    tags:
```
danbudris, line 4: I think we can remove these tag triggers

danbudris (approved):

PR #780, opened 2023-01-19
Updated kops version and patch

kschumy:
*Issue #, if available:*

*Description of changes:*
* Updated to v1.26.0-beta.2. We will have to update it again when the non-beta version is out
* Updated patch to be able to apply to current version.
  * Also made changes so we don't have to update some insignificant things in the patch every time we update it


projects/kubernetes/kops/patches/0001-remove-hardcoded-requierment-on-flannel-plugin.patch
```diff
@@ -1,28 +1,26 @@
-From 2ec5afbedada5688c13aa8e8d9ac9ffba3ae4511 Mon Sep 17 00:00:00 2001
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
TerryHowe, line 1: Is this allowed? Like a wildcard or something?

kschumy: We do it off the EKS-D Kubernetes patches

PR #810, opened 2023-02-07
Support for building EKS-D in codebuild

danbudris:
*Issue #, if available:*
https://github.com/aws/eks-distro-internal/issues/269

*Description of changes:*
PR with the changes required to support EKS-D builds in Codebuild.

Includes three [buildspecs](https://docs.aws.amazon.com/codebuild/latest/userguide/build-spec-ref.html), one to build EKS Go for each supported version (projects/golang/go/build/conformanceTests/golangBuildspec.yaml), one to build the builder-base using those built Golang versions (projects/golang/go/build/conformanceTests/builderBaseRebuildBuildspec.yaml), and one to build EKS Distro (projects/golang/go/build/conformanceTests/eksDistroRebuildBuildspec.yaml) using the re-built builder-base image.

The Golang and EKS-D buildspecs use [batch builds](https://docs.aws.amazon.com/codebuild/latest/userguide/batch-build.html) to run parallel builds for each supported Golang and K8s version. The artifacts from the Golang build (e.g. the RPMs) are aggregated at the end of the build step and all of the built artifacts are supplied to the build base build step, so we end up with a builder-base which has a custom-built version for each supported Golang version. The EKS-D rebuild buildspec runs the EKS-D build and conformance test execution for each supported K8s version, running in the custom builder-base from the previous step.

We also add utility scripts for setting up the ECR credentials helper (required for pushing to ECR), initializing an uninitialized Git repository (which is required as the codepipeline -> Github integration copies the source as a dir structure not as a cloned git repo, and we use git repo actions throughout), other common functionality like `retry` and the Docker pull retry, and a utility script to set up the [Qemu](https://www.qemu.org/) emulator on the codebuild host for ARM builds.

This PR is also dependent on its counterpart in EKS Distro -- https://github.com/aws/eks-distro/pull/1732


projects/golang/go/build/conformanceTests/eksDistroRebuildBuildspec.yaml
```diff
@@ -0,0 +1,66 @@
...
+    SKIP_CHECKSUM_VALIDATION: "true"
+    TARGET_BRANCH: "codebuildTest"
+    TEST_ROLE_ARN: "arn:aws:iam::125833916567:role/TestBuildRole"
+
+phases:
+  pre_build:
+    commands:
+      - echo Running pre-build setup and initialization...
+      - scripts/setup_ecr_credential_helper.sh
+      - scripts/setup_qemu.sh
```
markapruett, line 61: Is there no arm option when running on codebuild?

projects/golang/go/build/conformanceTests/golangBuildspec.yaml
```diff
@@ -0,0 +1,66 @@
...
+        variables:
+          GO_SOURCE_VERSION: 1.18
+    - identifier: golang_119
+      env:
+        variables:
+          GO_SOURCE_VERSION: 1.19
+    - identifier: golang_120
+      env:
+        variables:
+          GO_SOURCE_VERSION: 1.20
```
markapruett, line 44: Once this merges we should add this to the runbook as another place to update when making changes/adding a new minor version.

builder-base/Makefile
```diff
@@ -81,7 +83,7 @@ images-minimal: FINAL_STAGE_BASE=minimal-copy-stage
 images-minimal: VARIANT=minimal
 
 .PHONY: images-%
-images-%: buildkit-check download-golang
+images-%:
```
jaxesn, line 86: did u put the buildkit-check call in the prow job yaml?

projects/golang/go/build/conformanceTests/eksDistroRebuildBuildspec.yaml
jaxesn: there is but youd have to separate the build out across two instances/jobs then "merge" back together which for this job is probably unnecessary, most of the image builds are pretty simple and do not even trigger emulation.  As an example, we do use arm instances for one of the builds in the eks-a [staging](https://github.com/aws/eks-anywhere-build-tooling/blob/main/release/staging-build.yml#L182) build.

jaxesn (commented):
Some minor comments/questions, but this looks pretty good to me.

builder-base/Makefile
danbudris: yeah these are running as a pre-req in the prow jobs https://github.com/aws/eks-distro-prow-jobs/blob/main/jobs/aws/eks-distro-build-tooling/builder-base-postsubmits.yaml#L62

projects/golang/go/build/conformanceTests/eksDistroRebuildBuildspec.yaml
```diff
@@ -0,0 +1,66 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This file was generated by running `make generate`. DO NOT EDIT
+
+version: 0.2
+
+batch:
+  fast-fail: true
+  build-graph:
```
kschumy, line 21: just curious but is there a reason why we don't include 1.26?

markapruett (approved):

PR #823, opened 2023-02-09
Add git image for ops tool

rcrozean:
*Issue #, if available:*

*Description of changes:*
Output logs from the ops tool failing from `\"git\": executable file not found in $PATH",`

### DIFF
--- a/tools/eksDistroBuildToolingOpsTools/docker/Dockerfile
+++ b/tools/eksDistroBuildToolingOpsTools/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM public.ecr.aws/eks-distro-build-tooling/eks-distro-base:latest
+FROM public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-git:latest-al2
 
 ARG BINARY_OUTPUT_PATH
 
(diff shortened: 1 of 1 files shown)


tools/eksDistroBuildToolingOpsTools/docker/Dockerfile
```diff
@@ -9,4 +9,6 @@ ADD $BINARY_OUTPUT_PATH /usr/bin
 
 RUN chmod +x /usr/bin/$BINARY_NAME
```
jaxesn, line 10: Why does this need the chmod? isnt this a built go binary?

tools/eksDistroBuildToolingOpsTools/docker/Dockerfile
```diff
@@ -9,4 +9,6 @@ ADD $BINARY_OUTPUT_PATH /usr/bin
...
 
+RUN yum install git -y
```
jaxesn, line 12: As a matter of best practice, when yum installing in dockerfiles, clean up afterwards...

https://github.com/aws/eks-distro-build-tooling/blob/main/eks-distro-base/Dockerfile.base#L26-L27

jaxesn (changes requested):

tools/eksDistroBuildToolingOpsTools/docker/Dockerfile
jaxesn: Id also use `git-core` instead of `git`. git includes a number of extra deps that you most likely dont need in a container image like this

tools/eksDistroBuildToolingOpsTools/docker/Dockerfile
rcrozean: That is a god question. It is the binary so I don't think we need it which could mean we use the git minimal image, right?

jaxesn (approved):

PR #827, opened 2023-02-10
Scaffolding for EKS D consumer updater

danbudris:
*Issue #, if available:*

*Description of changes:*
A quick pass at setting up a structure for the EKS D consumer updater, or 'downstream automatin', in the EKS-D ops tool. We should consider porting over @kschumy's work to automation the bottlerocket update into this structure, and we can add implementations of the consumer updater for other projects that are downstream of us


rcrozean:
Is there a need for separate binaries or would it make sense to have a single binary? Something like 'eksdOpsTool update customer -n <cu stomerName>'

danbudris:
> Is there a need for separate binaries or would it make sense to have a single binary? Something like 'eksdOpsTool update customer -n '

yeah, I don't feel strongly either way -- it'd be fine to have this as a sub-command in the main ops tool

PR #861, opened 2023-02-17
Generate Patches for Go 1.17

zafs23:
*Issue #, if available:*

*Description of changes:*


projects/golang/go/1.17/rpmbuild/SPECS/golang.spec
```diff
@@ -538,6 +542,12 @@ fi
 %endif
 
 %changelog
+* Wed Feb 17 2023 Sajia Zafreen <szafreen@amazon.com> - 1.17.13-3
```
rcrozean, line 545: Need to add the correct date. If it was Wed then it should read `Wed Feb 15 2023` it has fail during the build process if this date is incorrect.

PR #865, opened 2023-02-21
unifiy golang and debian base image prod releases

danbudris:
*Issue #, if available:*
https://github.com/orgs/aws/projects/23/views/26?pane=issue&itemId=18822067

*Description of changes:*
Now, when we release a Golang Debian Base Image, the tags will:
- include one tagged simply with the latest Git tag for the given version, with no other identifiers, to act like a 'latest' image
- include a tag which includes the Git tag, the EKS Go release number, and the Dockerfile release number

The Debian dockerfile release will also take place as a part of the standard EKS Go prod release. This way, we can still update the Dockerfile independently of the EKS Go release process in order to trigger a release and modify the image, but we also get the convenience of automation release when we release a new prod version of EKS Go. See the ticket for the whole picture.

Requires some updates to the prow jobs, [specifically including an image build in each one.](https://github.com/aws/eks-distro-prow-jobs/pull/482)

### DIFF
--- a/projects/golang/go/Makefile
+++ b/projects/golang/go/Makefile
@@ -75,6 +75,7 @@ release: $(RELEASE_TARGETS)
 prod-release: golang-build sync-artifacts-to-s3
 
 .PHONY: prod-release-images
+prod-release-images: export AWS_PROFILE=ecr-public-push
 prod-release-images: fetch-eks-go-archive images
 
 .PHONY: fetch-eks-go-archive
--- a/projects/golang/go/Makefile
+++ b/projects/golang/go/Makefile
@@ -35,8 +35,9 @@ AWS_ACCOUNT_ID?=$(shell aws sts get-caller-identity --query Account --output tex
 # This build is specifically focused on a debian base image for use with EKS build systems
 IMAGE_REPO?=$(AWS_ACCOUNT_ID).dkr.ecr.$(AWS_REGION).amazonaws.com
 IMAGE_NAME?=golang-debian
-IMAGE_TAG?=$(GIT_TAG)-$(IMAGE_BUILD_ID)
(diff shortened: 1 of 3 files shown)


projects/golang/go/Makefile
```diff
@@ -35,8 +35,9 @@ AWS_ACCOUNT_ID?=$(shell aws sts get-caller-identity --query Account --output tex
 # This build is specifically focused on a debian base image for use with EKS build systems
 IMAGE_REPO?=$(AWS_ACCOUNT_ID).dkr.ecr.$(AWS_REGION).amazonaws.com
 IMAGE_NAME?=golang-debian
-IMAGE_TAG?=$(GIT_TAG)-$(IMAGE_BUILD_ID)
-IMAGE?=$(IMAGE_REPO)/$(IMAGE_NAME):$(IMAGE_TAG)
+IMAGE_TAG?=$(GIT_TAG)-$(BUILD_ID)-$(IMAGE_BUILD_ID)
+LATEST_IMAGE=$(IMAGE_REPO)/$(IMAGE_NAME):$(GIT_TAG)
+IMAGE?=$(IMAGE_REPO)/$(IMAGE_NAME):$(IMAGE_TAG),$(LATEST_IMAGE)
```
markapruett, line 40: Will this change the versioning string for all versions (not just the debian one?). Looking at hte makefile, I'm unclear how this gets used.

danbudris: so we only use this specific makefile to build the debian images; it'll change for that only

danbudris: so moving forward, any of the EKS Go debian base images will have the new image tag schema

danbudris: this is used only as an input to `buildkit` - the `--output` arg in the image build is passed this as the name/tags of the images it should output

markapruett (approved):

PR #867, opened 2023-02-22
Bottlerocket consumer updater ops tool

danbudris:
*Issue #, if available:*

*Description of changes:*
Adds a consumer updater for bottlerocket -- currently just ported over from the existing upater, and uses the local path. We will move this to use in-memory git later

Add a manifest fetcher to the EKS Distro release object in the Ops tool. This makes use of the Release types defined in `cmd/release` to marshal the Release Manifest JSON into a Go struct and make it available via the Release object. 

This can be used to fetch an existing release manifest for a given EKS D version; the functionality will be used in a follow-on PR to determine wether or not we need to invoke the BottleRocket Updater, based on the SHA of the K8s src manifest compared to the SHA in the BR source.

Much of this is based on the work by @kschumy for the bottlerocket updater


tools/eksDistroBuildToolingOpsTools/pkg/consumerUpdater/bottlerocket.go
```diff
@@ -1,13 +1,24 @@
...
+	"path/filepath"
+	"regexp"
 
 	"github.com/aws/eks-distro-build-tooling/tools/eksDistroBuildToolingOpsTools/pkg/eksDistroRelease"
 )
 
 const (
-	bottlerocketName = "Bottlerocket"
+	bottlerocketName    = "Bottlerocket"
+	versionRegex        = `[0-9]+\.[0-9]+\.[0-9]+`
```
rcrozean, line 16: This will be a very consistent pattern do we want to move it to pkg/constants.go.

tools/eksDistroBuildToolingOpsTools/pkg/consumerUpdater/bottlerocket.go
```diff
@@ -1,13 +1,24 @@
...
+	filePermission      = 0644
```
rcrozean, line 17: For file permissions, do we want to give a name or description of the permissions given and moved to pkg/constants.go as well?

tools/eksDistroBuildToolingOpsTools/pkg/consumerUpdater/bottlerocket.go
danbudris: done

tools/eksDistroBuildToolingOpsTools/pkg/consumerUpdater/bottlerocket.go
danbudris: done

PR #872, opened 2023-02-23
Git client for EKS-D Ops Tool

danbudris:
*Issue #, if available:*
https://github.com/aws/eks-distro-build-tooling/issues/876
https://github.com/aws/eks-distro-build-tooling/issues/877

*Description of changes:*
Adding a full-fledged Git client to the EKS-A ops tool, which supports operations like cloning into memory, adding and removing files, creating remotes, pushing, pulling, etc.

A lot of this is copied from my work on EKS-A GitOps providers, with modifications made to allow in-memory clones of repositories. 

Once this gets merged, I'm going to add this to the bottlerocket updater so that it clones the br repo into memory and makes the file changes in that specific dir before pushing to a remote for PR creation.We could also use the existing Github repo manager client to then create a PR.


tools/eksDistroBuildToolingOpsTools/pkg/git/gogitClient.go
```diff
@@ -0,0 +1,501 @@
...
+
+	logger.V(3).Info("Opening working tree")
+	w, err := g.Client.OpenWorktree(r)
+	if err != nil {
+		return err
+	}
+
+	logger.V(3).Info("Generating Commit object...")
+	commitSignature := &object.Signature{
+		Name: "EKS-A",
```
danbudris, line 150: this needs to be changed

PR #885, opened 2023-03-03
Adds iptables-wrapper to iptables minimal image

jaxesn:
*Issue #, if available:*

*Description of changes:*

This should resolve https://github.com/aws/eks-distro/issues/1729.

@g-gaston has been working with [upstream](https://github.com/kubernetes-sigs/iptables-wrappers/pull/6) on a golang based iptables-wrapper which we could use in our minimal images to properly switch the kube-proxy image from iptables-legacy to iptables-nft.  This PR pulls in the current state of this PR, which based on our testing seems pretty solid. This code is also very simple and easy enough to reason about that I feel comfortable getting a bit ahead here, esp considering our kube-proxy image does not work properly on hosts with iptables-nft as the default.

I have chosen to add the wrapper to the minimal iptables images, vs adding in the eks-d build so that we can utilize this in the various consumers of this base image (eks-d, eks minimal addon, eks-a kind image).  This change introduces the binary to the final along with a new set-alternative which *should* be fully backward compatible with previous versions of this image, allowing consumers to "opt-in" to use the new wrapper.  The default will continue to be iptables-legacy. 

Once merged and built, we can make a simple change to eks-d, basically just this [line](https://github.com/aws/eks-distro/pull/1728) to swap the default to the wrapper alternative, which should fix the legacy/nft issue.

/hold

### DIFF
--- a/eks-distro-base/Dockerfile.minimal-base-iptables
+++ b/eks-distro-base/Dockerfile.minimal-base-iptables
@@ -16,6 +16,18 @@
 # *NOTE* we have to limit our number of layers heres because in presubmits there
 # is no overlay fs and we will run out of space quickly
 
+################# IPTABLES WRAPPER BUILDER #####################
+
+ARG BUILDER_IMAGE
+ARG BASE_IMAGE
+
+FROM public.ecr.aws/eks-distro-build-tooling/golang:1.20 as iptables-wrapper-builder
+
+ADD iptables-wrappers/ /iptables-wrappers
+RUN set -x && \
+    cd /iptables-wrappers && \
+    make build
+
 ################# BUILDER #####################
 ARG BASE_IMAGE=unused
 ARG BUILDER_IMAGE=unused
@@ -28,6 +40,8 @@ ARG OUTPUT_DEBUG_LOG
 # built it has the latest scripts in the builder
 COPY scripts/ /usr/bin
 
+COPY --from=iptables-wrapper-builder /iptables-wrappers/bin/iptables-wrapper /newroot/usr/sbin
+
 RUN set -x && \
     export OUTPUT_DEBUG_LOG=${OUTPUT_DEBUG_LOG} && \
     if_al2 install_rpm chkconfig && \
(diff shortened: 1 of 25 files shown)


jaxesn:
Tested in a kops using a custom build of kube-proxy on both Ubuntu 22.04 and 20.04 and the wrapper correctly picked the mode.

eks-distro-base/Dockerfile.minimal-base-iptables
```diff
@@ -38,15 +52,29 @@ RUN set -x && \
...
     # https://github.com/kubernetes/release/blob/master/images/build/debian-iptables/buster/iptables-wrapper
     for m in nft legacy; do chroot $NEWROOT update-alternatives --install /usr/sbin/ip6tables ip6tables /usr/sbin/ip6tables-$m 10 \
             --slave /usr/sbin/ip6tables-restore ip6tables-restore /usr/sbin/ip6tables-$m-restore \
             --slave /usr/sbin/ip6tables-save ip6tables-save /usr/sbin/ip6tables-$m-save; done && \
-    # ensure iptables are in legacy mode
+    # The newer version of the iptables-wrapper.sh: https://github.com/kubernetes-sigs/iptables-wrappers/blob/master/iptables-wrapper-installer.sh
+    # And the in-progress golang version supports RHEL's configuration of alternatives, where iptables and ip6tables is managed via the same record
+    # We are now including the built golang based iptables-wrapper as a possible alternative for downstream consumers (eks-d kube-proxy, eks addon kube-proxy, eks-a kind image)
+    # Since this verison supports a single alternative record, we are adding the iptables-wrapper as a single record which is different than the above which is left for backwards compat
```
g-gaston, line 69:
```suggestion
    # Since this version supports a single alternative record, we are adding the iptables-wrapper as a single record which is different than the above which is left for backwards compat
```

eks-distro-base/Dockerfile.minimal-base-iptables
```diff
@@ -38,15 +52,29 @@ RUN set -x && \
...
+    # Not setting it as the default, but intend to in the future once all of our known consumers have decided to take this as their default mode
+    chroot $NEWROOT update-alternatives --install /usr/sbin/iptables iptables /usr/sbin/iptables-wrapper 100 \
+        --slave /usr/sbin/iptables-restore iptables-restore /usr/sbin/iptables-wrapper \
+        --slave /usr/sbin/iptables-save iptables-save /usr/sbin/iptables-wrapper \
+        --slave /usr/sbin/ip6tables ip6tables /usr/sbin/iptables-wrapper \
+        --slave /usr/sbin/ip6tables-restore ip6tables-restore /usr/sbin/iptables-wrapper \
+        --slave /usr/sbin/ip6tables-save ip6tables-save /usr/sbin/iptables-wrapper && \
+    # default to iptables in legacy mode
     for t in iptables ip6tables; do chroot $NEWROOT update-alternatives --set $t /usr/sbin/$t-legacy; done && \
```
g-gaston, line 78: So if I'm getting this correctly, we install the golang iptables-wrapper with `update-alternatives` but line 78 updates the symlinks back to the legacy binaries, right?

So it's up to each child image consumer of this image to update the symlinks again if they want to use the wrapper. Assuming that they would just do something like like 78 but pointing to the wrapper?

eks-distro-base/Dockerfile.minimal-base-iptables
```diff
@@ -38,15 +52,29 @@ RUN set -x && \
...
-    # AL2 doesn't setup the ip6tables alternative the same as ubuntu/debian
-    # which is what the iptables-wrapper script provided by kubernetes expects
-    # since kind's entrypoint assumes this adding the alternative
-    # adding the alternative for ip6tabels here so the --set calls would work later on as expected
+    # The original iptables-wrapper.sh script assumed that there were both iptables and ip6tables 
+    # alternatives setup since the upstream image was based on debian which sets them up that way
+    # AL2 follows RHEL which uses a single alternaive setup for both iptables and ip6tables
+    # The wrapper script is currently only used by the eks-a kind image, since the other consumers of this base image
+    # do not include bash and have therefore relied on legacy as the default
+    # This adds the ip6tables atlernative so that the `update-alternatives --set` calls which happen during the kind
```
g-gaston, line 60:
```suggestion
    # This adds the ip6tables alternative so that the `update-alternatives --set` calls which happen during the kind
```

eks-distro-base/tests/run_tests.sh
```diff
@@ -163,22 +219,118 @@ function check_base-iptables() {
...
+                if ! docker run --privileged --net host --rm --platform=$platform $LOCAL_REGISTRY/eks-distro-minimal-images-base-test:iptables-wrapper-with-bash-latest bash -c "$ip --version && update-alternatives --list" | grep "ip6tables	manual	/usr/sbin/ip6tables-$mode"; then
+                    echo "ip6tables wrapper alternative issue!"
+                    exit 1
+                fi
+            done
+        done
+        
+        cleanup-iptable-rules
+
+        # run upstream test binaries which tests a number of the same cases as above
```
g-gaston, line 314: is anything that you are testing above this line missing from the go tests?
If so, let me and I'll add it.

eks-distro-base/Dockerfile.minimal-base-iptables
jaxesn: yeah exactly. the wrapper will be avail but the default will be left the same, legacy.

jaxesn:
The null in the tag file breaks the kind presubmit.

/override eks-distro-base-tooling-presubmits-kind-al-2 eks-distro-base-tooling-presubmits-kind-al-2022

markapruett (approved):

PR #894, opened 2023-03-07
Add pushgateway for metrics

rcrozean:
*Issue #, if available:*

*Description of changes:*
Adds the pushgateway to the deployments for prow. This is where the ghproxy and other portions of Prow use this as a single scrape point for metrics. Details in doc https://docs.prow.k8s.io/docs/metrics/


helm-charts/stable/prow-control-plane/templates/pushgateway-Deployment.yaml
```diff
@@ -0,0 +1,113 @@
...
+    metadata:
+      labels:
+        app: pushgateway-proxy
+    spec:
+      containers:
+      - name: nginx
+        ports:
+          - name: http
+            containerPort: 8081
+        image: nginx:1
```
abhay-krishna, line 90: Let's use nginx from ECR Public here.

rcrozean: Should we break this out into the values.yaml or are we fine leaving it here

abhay-krishna: We should put all configurable values in values.yaml

PR #896, opened 2023-03-08
Update schema for ingress api v1

rcrozean:
*Issue #, if available:*

*Description of changes:*
These changes are needed for deployment. The schema changed with the move to networking.k8s.io/v1. These changes were based off the upstream deployment yaml -> [ingress.yaml](https://github.com/kubernetes/test-infra/blob/master/config/prow/cluster/ing_ingress.yaml)


helm-charts/stable/prow-control-plane/templates/deck-Ingress.yaml
```diff
@@ -16,27 +16,36 @@
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
+  name: prow
+  namespace: default
```
abhay-krishna, line 20: nit: can be removed

abhay-krishna (approved):
/approve

PR #898, opened 2023-03-09
Update to Go 1.19.7

kschumy:
*Issue #, if available:*

*Description of changes:*


projects/golang/go/1.19/patches/0102-syscall-expose-IfInfomsg.X__ifi_pad-on-s390x.patch
```diff
@@ -1,4 +1,4 @@
-From 67a4711d09c6595c17f32470c15bf471c287777d Mon Sep 17 00:00:00 2001
```
markapruett, line 1: Did these patches need to be regenerated?

PR #912, opened 2023-03-14
Changes al22 to al23 to match new al version

jaxesn:
*Issue #, if available:*

*Description of changes:*

This is mostly just a find a replace for 2022->2023. There are a couple real changes I will call out based on changes to 2023 that we did not see in 2022.

Note: most of the 2023 builds are going to fail because they depend on images already being pushed. I have built everything locally and ran a previous run with some specific hardcodes to ensure they all built in prow.


eks-distro-base/Dockerfile.minimal-base
```diff
@@ -39,6 +39,13 @@ RUN set -x && \
         findutils \
         shadow-utils \
         yum-utils && \
+    if grep -q "2023" "/etc/os-release"; then \
```
jaxesn, line 42: documented the oddity here

eks-distro-base/scripts/install_binary
```diff
@@ -70,6 +70,7 @@ function build::install::binary() {
     # we need to remove the link before copying
     [ -L $final_bin ] && unlink $final_bin
 
+    mkdir -p $(dirname $final_bin)
```
jaxesn, line 73: surprised we hadnt seen this before, but this just ensures the dest folder exists before trying to copy.

markapruett (approved):

PR #1025, opened 2023-06-07
Generate EKS-Go 1.17.13 patches

zafs23:
*Issue #1020 , if available:*

*Description of changes:* We are not moving forward with this go patches as EKS-D has decided to bump all the projects using EKS-Go 1.17 to EKS-Go 1.18+


zafs23:
Bumped all projects using go1.17 to go1.18+

PR #1048, opened 2023-06-28
changes iptables minimal image default to use iptables-wrapper

jaxesn:
*Issue #, if available:*

*Description of changes:*

We added the golang based [iptables-wrapper ](https://github.com/kubernetes-sigs/iptables-wrappers/blob/master/main.go)script to the minimal iptables image a few months ago. At the time we decided to leave the default "alternative" as iptables-legacy to avoid casually breaking downstream consumers of this image.  At the time we updated eks-distro's [kube-proxy](https://github.com/aws/eks-distro/blob/main/projects/kubernetes/release/docker/kube-proxy-base/Dockerfile) to use the new wrapper.

We are now ready to make the wrapper the default alternative to make it easier for downstream consumers to build their images to rely on this wrapper to properly flip between legacy/nft iptables mode based on the host OS.

*Notes*

After this merges and is buiilt, the `latest` tag will point to the new image with the wrapper set as the default alternative. This means the next time any consumers build their images they will get one with this default.  This also means there will be no more cve updates to the iptables image with the previous default.  The eks-distro build could change its dockerfile, but does not have to since its already setting the same alternative.  I have spoken with all the internal consumers I am aware of and made sure they are prepared to take this change.

We also believe this change should not negatively impact anyone using this image, in fact it will probably behave more like they expect, but have not noticed since only recently OS's started shipping with iptables-nft as the default (rhel 8.4+).

ref: https://github.com/aws/eks-distro/issues/2160

### DIFF
--- a/eks-distro-base/Dockerfile.minimal-base-iptables
+++ b/eks-distro-base/Dockerfile.minimal-base-iptables
@@ -74,8 +74,8 @@ RUN set -x && \
         --slave /usr/sbin/ip6tables ip6tables /usr/sbin/iptables-wrapper \
         --slave /usr/sbin/ip6tables-restore ip6tables-restore /usr/sbin/iptables-wrapper \
         --slave /usr/sbin/ip6tables-save ip6tables-save /usr/sbin/iptables-wrapper && \
-    # default to iptables in legacy mode
-    for t in iptables ip6tables; do chroot $NEWROOT update-alternatives --set $t /usr/sbin/$t-legacy; done && \
+    # default to iptables to use golang wrapper
+    chroot $NEWROOT update-alternatives --set iptables /usr/sbin/iptables-wrapper && \
     # Remove bad symlinks due to deleted man files
     find $NEWROOT/etc/alternatives -xtype l -name "*-man" -delete -print && \
     cleanup "iptables"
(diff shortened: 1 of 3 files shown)


eks-distro-base/tests/Dockerfile
```diff
@@ -28,6 +28,9 @@ CMD ["/bin/check-cgo"]
 
 FROM ${BASE_IMAGE} as check-iptables-legacy
 
+RUN ["update-alternatives", "--set", "iptables", "/usr/sbin/iptables-legacy"]
```
jaxesn, line 31: this is a dockerfile we use to run some basic docker tests with these images. I changed the legacy target to explicitly set the alternative and removed it from the wrapper target.

eks-distro-base/Dockerfile.minimal-base-iptables
```diff
@@ -74,8 +74,8 @@ RUN set -x && \
         --slave /usr/sbin/ip6tables ip6tables /usr/sbin/iptables-wrapper \
         --slave /usr/sbin/ip6tables-restore ip6tables-restore /usr/sbin/iptables-wrapper \
         --slave /usr/sbin/ip6tables-save ip6tables-save /usr/sbin/iptables-wrapper && \
-    # default to iptables in legacy mode
-    for t in iptables ip6tables; do chroot $NEWROOT update-alternatives --set $t /usr/sbin/$t-legacy; done && \
+    # default to iptables to use golang wrapper
```
danbudris, line 77:
```suggestion
    # default to iptables using golang wrapper
```

jaxesn:
/override eks-distro-base-tooling-presubmits-kind-al-2023 eks-distro-base-tooling-presubmits-kind-al-2

markapruett (approved):

PR #1072, opened 2023-07-18
Add architecture argument for golang_download script

zafs23:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/builder-base/scripts/download_golang.sh
+++ b/builder-base/scripts/download_golang.sh
@@ -61,8 +61,6 @@ function build::go::download(){
     done
 }
 
-if 
-
 if [[ $ARCHITECTURE == "linux/amd64" ]] ; then 
     build::go::download "${VERSION}" "$OUTPUT_DIR" "x86_64"
 fi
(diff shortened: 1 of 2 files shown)


builder-base/scripts/download_golang.sh
```diff
@@ -60,5 +61,10 @@ function build::go::download(){
     done
 }
 
-build::go::download "${VERSION}" "$OUTPUT_DIR" "x86_64"
-build::go::download "${VERSION}" "$OUTPUT_DIR" "aarch64"
+if [[ $( echo $ARCHITECTURE | cut -d ',' -f1 ) == "linux/amd64" ]] ; then 
+    build::go::download "${VERSION}" "$OUTPUT_DIR" "x86_64"
+fi
+
+if [[ $( echo $ARCHITECTURE | cut -d ',' -f2 ) == "linux/arm64" ]] ; then
```
danbudris, line 68: hm this assumes that the`linux/amd64, linux/arm64` are always in the same order right? i think that'll bite us at some point in the future, can we do this with a 'find' or some other thing that doesn't care about the order?

danbudris (approved):

PR #1091, opened 2023-08-02
WIP: Update the ops tool to auto cherry pick on backport requests

rcrozean:
*Issue #, if available:*

*Description of changes:*
These changes are to take the current chatops command `/backport:go <versions>` and attempt to auto cherrypick the commit upstream closed their ticket with.

The changes to `server_test.go` are to mock the newly added calls in server.

There are a few additional updates that need to happen like writing the functions to `exec.command` locally to run `cherry-pick` and `format-patch` since the prow client doesn't provide those. We could add these upstream, but that would require waiting for the merge, then updating the prow cluster to the commit of prow with the new commands when they make their new image. So for the time being going to add them here.

In `golangPatchReleaseRequest` I don't think we will ever want to allow this to be triggered by anyone outside the AWS org or our bot, so I removed the check to allow anyone to trigger and just check for AWS org or bot.


tools/eksDistroBuildToolingOpsTools/pkg/externalplugin/backportGolangRequest.go
kschumy, line 1: This function is pretty long. Can we break it into separate functions if possible?

PR #1139, opened 2023-08-30
Copy check updates for quarterly check updates

rcrozean:
*Issue #, if available:*

*Description of changes:*
Instead of rebuilding the minimal images every quarter regardless of the requirements to the images, we check the images for any updates the packages may require similar to how we check for security updates.


eks-distro-base/Makefile
```diff
@@ -746,6 +752,7 @@ test-all-minimal-images: $(addprefix test-minimal-images-, $(MINIMAL_VARIANTS))
 standard-update: UPDATE_TARGET=standard-images
 standard-update: IMAGE_NAME=eks-distro-base
 standard-update: IMAGE_COMPONENT=$(IMAGE_NAME)
+standard-update: SECURITY?=
```
jaxesn, line 755: I dont *think* you want/need this here since this target is called as a part of the `update` target which will it accordingly.

jaxesn (commented):
This looks pretty good! One comment which I am not 100% sure on.  Id probably try to test this somehow, maybe add `echo` in front of the actual builtkit command so you can run it quickly and just verify what it will do in different cases?

eks-distro-base/Makefile
rcrozean: For some reason before this was added we were getting [test failures](https://prow.eks.amazonaws.com/view/s3/prowpresubmitsdataclusterstack-prowbucket7c73355c-vfwwxd2eb4gp/pr-logs/pull/aws_eks-distro-build-tooling/1139/eks-distro-base-test-presubmit/1735043820032102400)

eks-distro-base/Makefile
```diff
@@ -770,14 +772,15 @@ release: buildkit-check $(IMAGE_TARGETS)
 create-pr: $(CREATE_PR_TARGETS)
 	$(MAKE_ROOT)/../pr-scripts/create_pr.sh eks-distro-build-tooling 'EKS_DISTRO*_TAG_FILE*' $(IMAGE_UPDATE_BRANCH)
 
-# Used in the quarterly minimal image rebuild prowjob. It adapted the existing create_pr script.
-.PHONY: create-rebuild-pr
-create-rebuild-pr: IMAGE_REBUILD_BRANCH=rebuild-all-minimal-images
-create-rebuild-pr: 
-	$(MAKE_ROOT)/../pr-scripts/create_pr.sh eks-distro-build-tooling 'EKS_DISTRO_TAG_FILE' $(IMAGE_REBUILD_BRANCH)
+.PHONY: echo
```
jaxesn, line 775: is this needed?

eks-distro-base/Makefile
```diff
@@ -770,14 +772,15 @@ release: buildkit-check $(IMAGE_TARGETS)
...
+echo:
 
 .PHONY: update
-update: buildkit-check open-pr-check $(UPDATE_TARGETS)
+update: SECURITY?=--security
+update: var-value-SECURITY \
```
jaxesn, line 780: why `var-value-SECURITY` as a prereq?

rcrozean: Left over from testing. Will remove.

jaxesn (approved):

PR #1210, opened 2023-10-13
installs python 3.9 from minimal base image into builder base

jaxesn:
*Issue #, if available:*

*Description of changes:*

For the later versions of ansible, which is only used by our image-builder builds, we need a python 3.9 version. AL2 does not ship a 3.9 version by default.  When we originally built the emissary image, which also needed 3.9, we built a al2 based minimal python image building python from source.

This PR copies python 3.9 from the minimal image into the builder base instead of using 3.8 from yum.  This *shouldnt* affect any of the builds since we are pretty sure the only build relying on python3 is the image-builder builds.


builder-base/Dockerfile
```diff
@@ -194,7 +195,29 @@ RUN --mount=type=cache,target=/var/cache/yum,sharing=locked \
...
+COPY --link --from=python-3.9 / /ansible
+# we want python2 to be the version symlinkd to python
+# we also do not want to overwrite the rpm db
+# this will leave the packages pulled in from the python image in
+# a slightly odd state in that they are installed yet not in the rpm db
+RUN set -x && \
+    unlink /ansible/usr/bin/python && \
+    rm -rf /var/lib/
+
+FROM ${BUILDER_IMAGE} as ansible-al23
```
jaxesn, line 220: for the al23 base image continue doing what we were since that version includes python 3.9

builder-base/Makefile
```diff
@@ -46,14 +47,16 @@ IGNORE_NO_CACHE?=false
...
-$(shell yq e ".al$(AL_TAG).$(1)" $(MAKE_ROOT)/../EKS_DISTRO_TAG_FILE.yaml)
+$(shell yq e ".al$(AL_TAG).\"$(1)\"" $(MAKE_ROOT)/../EKS_DISTRO_TAG_FILE.yaml)
 endef
 
 define NEWLINE
 
 
 endef
 
+REALPATH=TZ=utc $(shell if [ "$$(uname -s)" = "Darwin" ] && command -v grealpath &> /dev/null; then echo grealpath; else echo realpath; fi)
```
jaxesn, line 58: fix to run on mac

markapruett (approved):

PR #1284, opened 2023-12-12
Update with common mac os find

rcrozean:
*Issue #, if available:*

*Description of changes:*
Add a common function like https://github.com/aws/eks-anywhere-build-tooling/commit/e24cd141b68e8dc619f61ab07420273bf363a374#diff-a5ed650d97e91497a26f14a346a388676aba4e377ea1efff495154c860b3171b to find local commands for macOS.


builder-base/scripts/install_base_yum_packages.sh
```diff
@@ -19,29 +19,32 @@ set -o pipefail
 SCRIPT_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
 source $SCRIPT_ROOT/common_vars.sh
 
+BASE_DIRECTORY="$(git rev-parse --show-toplevel)"
```
jaxesn, line 22: Id probably skip this file since youll need to change the dockerfile for all the layers to include the new common.sh file.

PR #1317, opened 2024-01-19
Add oras to builder-base

panktishah26:
*Description of changes:*
Skopeo tool doesn't support newer attributes introduced by AWS Signer/notation. Thus, we need to use oras to copy container images from one repository to another along with signatures. Adding oras to builder-base.


jaxesn (commented):
This looks pretty great!

PR #1450, opened 2024-06-17
Update go install for minimal images to use the existing go version

rcrozean:
*Issue #, if available:*

*Description of changes:*
Since EKS Distro is deprecating EKS Go and don't want to continue building the RPMs for the versions upstream is supporting, we can use the existing golang installed on the builder base to install the newest versions. In this pr I am preserving the old eksgo install function but renaming them to `::eksgo::*`. The new function calls `go install` and `go download` to install and unpack the newest version of golang produced by upstream following [managing multiple golang installs](https://go.dev/doc/manage-install#installing-multiple). This followed with a prowjob to auto bump the `versions.yaml` and `null` the golang minimal images (https://github.com/aws/eks-distro-build-tooling/pull/1449) is the proposed new method for installing golang on the builder base


builder-base/scripts/install_golang.sh
```diff
@@ -30,68 +30,102 @@ source $SCRIPT_ROOT/common_vars.sh
...
-    mv /tmp/go-extracted/usr/lib/golang/* /root/sdk/$golang_version
+  version=$(echo "$golang_version" | grep -o "[0-9].*")
+  ln -s /root/sdk/go${version}/bin/go ${GOPATH}/bin/$golang_version
 
-    for license_dir in "/usr/share/licenses/golang" "/usr/share/doc/golang-$VERSION"; do
-        if [ -d /tmp/go-extracted/$license_dir ]; then
-            mv /tmp/go-extracted/$license_dir/* /root/sdk/$golang_version
-        fi
-    done
+  # This is only for 1.21+ and is causing failures.
```
jaxesn, line 71: id probably jsut remove this completely here since its not relevant to 1.21+ versions and its preserved down below for older versions.

builder-base/scripts/download_golang.sh
```diff
@@ -27,44 +27,63 @@ RELEASE_NUMBER="$(echo $VERSION | cut -d'-' -f 2)"
...
+}
 
-if [[ $ARCHITECTURE =~ "linux/arm64" ]] ; then 
-    build::go::download "${VERSION}" "$OUTPUT_DIR" "aarch64"
-fi
\ No newline at end of file
+if [[ ${VERSION:2:2} -ge "21" ]]; then
+  build::go::download "${VERSION}" "$OUTPUT_DIR" "$ARCHITECTURE"
+else
+  if [[ $ARCHITECTURE =~ "linux/amd64" ]]; then
```
jaxesn, line 82: id probably do something like [this](https://github.com/aws/eks-distro-build-tooling/blob/bd4d565dc6e2abeab70239def2d425ce5a7ad8f8/builder-base/scripts/install_govc.sh#L25) setting a temp var and only calling eksgo::download in one spot, but a not a huge deal.

builder-base/scripts/download_golang.sh
```diff
@@ -27,44 +27,63 @@ RELEASE_NUMBER="$(echo $VERSION | cut -d'-' -f 2)"
...
+  # Set up specific go version by using go get, additional versions apart from default can be installed by calling
+  # the function again with the specific parameter.
+  local version=${1%-*}
+  local outputDir=${2}
+  local archs=${3}
+
+  for arch in ${archs/,/ }; do
+    local filename="$outputDir/${arch}/go$version.${arch/\//-}.tar.gz"
+    if [ ! -f $filename ]; then
+      curl -sSLf --retry 5 "https://go.dev/dl/go$version.${arch/\//-}.tar.gz" -o $filename --create-dirs
```
jaxesn, line 74: it would probably be best to check the checksums here. if you want to follow the pattern we use elsewhere, we could check those in or download them from go.dev, like we did the rpm shas above.  again but not strictly required but would be at least slightly better

rcrozean: Made the additions for checking the sha and here is the local output.
```
+ curl -sSLf --retry 5 https://go.dev/dl/go1.22.4.linux-amd64.tar.gz -o /tmp/test/linux/amd64/go1.22.4.linux-amd64.tar.gz --create-dirs
++ curl -sSLf --retry 5 'https://go.dev/dl/?mode=json'
++ jq -r --arg tar go1.22.4.linux-amd64.tar.gz '.[].files[] | if .filename == $tar then .sha256 else "" end'
++ xargs
+ sha256sum=ba79d4526102575196273416239cca418a651e049c2b099f3159db85e7bade7d
++ sha256sum /tmp/test/linux/amd64/go1.22.4.linux-amd64.tar.gz
++ cut '-d ' -f1
+ [[ ba79d4526102575196273416239cca418a651e049c2b099f3159db85e7bade7d != \b\a\7\9\d\4\5\2\6\1\0\2\5\7\5\1\9\6\2\7\3\4\1\6\2\3\9\c\c\a\4\1\8\a\6\5\1\e\0\4\9\c\2\b\0\9\9\f\3\1\5\9\d\b\8\5\e\7\b\a\d\e\7\d ]]
```

jaxesn (approved):

PR #1515, opened 2024-08-23
if al2023, then install nftables

zachdorame:
*Issue #, if available:* N/A

*Description of changes:*

For supporting the nftables-based backend beta feature in EKS 1.31 for kube-proxy, we'll need to install ntables into the al2023 image used to build kube-proxy.
This PR is a *very* tiny change to allow the nftables binary to be installed

### DIFF
--- a/eks-distro-base/Dockerfile.minimal-base-iptables
+++ b/eks-distro-base/Dockerfile.minimal-base-iptables
@@ -53,7 +53,7 @@ RUN set -x && \
         kmod \
         iptables-nft && \
     if_al2 install_rpm iptables ebtables && \
-    if_al2023 install_rpm iptables-legacy ebtables-legacy && \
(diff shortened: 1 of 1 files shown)


jaydeokar:
@jaxesn @markapruett  
Would like to know your thoughts on this change. This is in for a beta feature that is going in v1.31.0. 

https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/3866-nftables-proxy/README.md
The requirement is that nftable version be greater than v1.0.1 which is only possible with AL2023 right now

zachdorame:
The current kube-proxy is based off the al2 base. However, we have evidence that there's no regressions in switching to al23 base as the al23 base kube-proxy passes our end-to-end tests.

jaxesn:
Makes sense to me! 🚢

markapruett (approved):

PR #1524, opened 2024-09-05
Add Go 1.23 to final stage in builder-base Dockerfile

abhay-krishna:
* Add Go 1.23 to final stage in builder-base Dockerfile so that it's available in the final output image
* Sort keys in `versions.yaml` tracker file, also add logic to the Golang version check script to handle this


builder-base/versions.yaml
```diff
@@ -8,23 +8,23 @@ GOLANG_VERSION_119: 1.19.13-14
 GOLANG_VERSION_120: 1.20.14-16
```
abhay-krishna, line 8: The changes here are purely cosmetic, just alphabetical sorting.

PR #1609, opened 2025-01-29
Trigger builder-base update for iptables

Ganiredi:
*Issue #, if available:*

*Description of changes:*

We have identified a CVE in kube-proxy, which uses the base image public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-iptables:latest-al23. To address this, we need to create a new image with an updated Go patch version.

image

Reproduce:

```
trivy image public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-iptables:latest-al23

trivy image public.ecr.aws/eks-distro/kubernetes/kube-proxy:v1.31.4-eks-1-31-latest

```


natechensan:
Can you add in the description what this change is doing and why it is OK?

PR #1768, opened 2025-08-05
 update 1-32 with patches require to support 1.34 k8s

Ganiredi:
*Issue #, if available:*

*Description of changes:*


vineeth-bandi:
instead of using a beta version should we just add a patch doing the equivalent?

projects/kubernetes/kops/GIT_TAG
```diff
@@ -1 +1 @@
-v1.32.1
+v1.33.0-beta.1
```
vineeth-bandi, line 1: instead of using a beta version should we just add a patch doing the equivalent?

PR #2003, opened 2026-06-08
Order iptables after golang-compiler and filter http2 noise

charles-large:
*Issue #, if available:*

*Description of changes:*

- Order iptables build after golang-compiler (Makefile + test expectations)
- Filter noisy http2 connection error logs from buildkit output (scripts/buildkit.sh)

When a new golang version is released, and EKS Distro Base Images build, iptables is being built before the golang images. This leads to iptables not using the latest go version. Created a dependency so that iptables will only build after the go version is built.

When checking the ordering locally.
```
printf "\n\nBuilding base-golang-compiler-1.25-base builder stage\n\n"
printf "\n\nBuilding base-golang-compiler-1.25-base final stage\n\n"
printf "Building base-golang-compiler-1.25-base complete!\n\n"
printf "\n\nBuilding base-iptables builder stage\n\n"
printf "\n\nBuilding base-iptables final stage\n\n"
printf "Building base-iptables complete!\n\n"
```

Context for the buildkit fix:

Go's HTTP/2 implementation logs http2: server connection error from ...: connection error: PROTOCOL_ERROR when a client reuses a connection the server has already sent GOAWAY for ([golang/go#32112](https://github.com/golang/go/issues/32112)). This is an informational diagnostic, not a failure. The HTTP/2 client recovers transparently by opening a new connection.

In Prow builds, the local registry sends GOAWAY under load after many sequential image builds. The docker buildx HTTP/2 client logs this to stderr, which gets merged into stdout via 2>&1. Since Makefile recipes interpret stdout as shell commands, the Go timestamp prefix (2026/06/08 20:10:30) gets executed by bash, causing bash: 2026/06/08: No such file or directory and failing the build despite the actual operation succeeding.

The fix filters this line from the build output, following the same pattern used in scripts/common.sh for suppressing noisy go-licenses messages.


charles-large:
duplicate https://github.com/aws/eks-distro-build-tooling/pull/1988

